PR #2, opened 2017-12-31
Add filter task-list

tarleb:
The filter adds support for GFM-style task-lists as discussed in jgm/pandoc#3051.

### DIFF
--- a/filters/task-list/task-list.lua
+++ b/filters/task-list/task-list.lua
@@ -1,0 +1,131 @@
+local List = require 'pandoc.List'
+
+local M = {}
+
+local function is_html (format)
+  return format == 'html' or format == 'html4' or format == 'html5'
+end
+
+--- Create a ballot box for the given output format.
+function M.ballot_box (format)
+  if is_html(format) then
+    return pandoc.RawInline(
+      'html',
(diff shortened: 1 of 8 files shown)


filters/task-list/task-list.lua
```diff
@@ -0,0 +1,58 @@
...
+            (blk.content[1].text == '[x]' or
+             blk.content[1].text == '[X]')) then
+    return ballot_box_with_check(), 1
+  else
+    return nil, 0
+  end
+end
+
+function ballot_box ()
+  if FORMAT == 'html' or FORMAT == 'html4' then
```
jgm: Do we also need `html5`?

jgm:
I've added you as a collaborator so you can merge these yourself.

Do you like the general layout of this repo?  If anything should be changed, in terms of expectations for additions, it will be easier to change it before too many filters are here!

tarleb:
Thanks!

I like the current layout, it seems well suited for our purposes.

It would be nice if we could reuse functions defined in filters or make filters defined here useable as modules (as to allow using them with `walk_*`), but I don't have a clear idea for that yet.

PR #5, opened 2018-01-07
Add section-refs filter.

jkr:
This allows section-by-section bibliopgraphies. It uses the output of
pandoc-citeproc and has to be run after it:

`pandoc -F pandoc-citeproc --lua-filter section-refs.lua`.

Essentially, it works as follow:

1. After pandoc-citeproc has generated a `refs` div, remove it.

2. After each section (designated by the appearance of (a) a section
heading of `$section-refs-level` or higher, or (b) the end of the
document) build a new sorted bibliography out of the removed refs div,
using the citations in that section. Insert it with in optional
header.


jkr:
Actually, I realize this uses `Citation.id` so it won't work with 2.0.6. @tarleb, is there a way to test pandoc-version for the filters version in 2.0.6? Just take the first line of `pandoc.pipe("pandoc", {"--version"}, "")`?

tarleb:
That seems like a good hack. Only realized today that this was missing 😐

tarleb:
Maybe using `io.popen` is a tiny bit cleaner than `pandoc.pipe`, but that's probably just a matter of taste.

tarleb:
Just realized: you could also test for the existence of `PANDOC_VERSION`. If it's non-nil, then it's pandoc 2.1, otherwise it's a pandoc 2.0.x.

jkr:
Oh, clever. I'll do that instead of the fragile pipe/gsub/tonumber hack I put in there.On Jan 7, 2018 12:11 PM, Albert Krewinkel <notifications@github.com> wrote:Just realized: you could also test for the existence of PANDOC_VERSION. If it's non-nil, then it's pandoc 2.1, otherwise it's a pandoc 2.0.x.

jkr:
Oh, duh -- this is still failing because it operates on pandoc-citeproc output, and so needs the filter-order fix that only comes in 2.1. So we don't really need the version check at all, I guess. I'll just wait till 2.1 to merge this.

tarleb:
I've given the `pandoc.utils.run_json_filter` another shot (jgm/pandoc#4227) and used this problem as a motivating example. However, I'm not sure I really understand the details. Could you take a look if the code snippet mentioned there would be a valid approach?

jkr:
I guess I'd worry about performance a bit -- as seen in an current discussion -discuss, citeproc can be slow things down quite a bit with a big bib (depending on the format). I guess you could keep your bib curated for the document, but I have to admit that's not what I do. So I'd be worried about parsing the thing each time.

Also -- I'm not sure I see how it determines which section level to put the bibs after. You don't want it after *each* sub-sub-sub-section (or maybe someone does, but others won't). I guess it could be customized in the `for _, sec in ipairs(sections)` loop by a conditional on the section number.

tarleb:
That point about performance is a good one, I hadn't considered that.  Maybe we could add a filter which automatically creates and updates the relevant bibliography subset.  That might actually be a better use-case for `run_json_filter` than what I came up with.

Indeed, the references section placement would need more work. Since I used `hierarchicalize` to make things simpler, the first level (as in `ipairs(sections)`) contains level-1 sections plus preceding blocks.  Currently each of those preceding blocks would get its own bibliography.  Supporting sub-section bibliographies would require a *walk*-like function for the output of hierarchicalize.

Thanks for your feedback!

jkr:
Now that 2.1 is out, I feel like this could be merged (at least as one possible solution for this problem). @tarleb do you have merging permission here?

tarleb:
Updated travis to use 2.1, should work now.

jgm:
+++ Albert Krewinkel [Jan 10 18 12:04 ]:
>   That point about performance is a good one, I hadn't considered that.
>   Maybe we could add a filter which automatically creates and updates the
>   relevant bibliography subset. That might actually be a better use-case

Note that pandoc-citeproc 0.13 automatically does this, so
working with large bibliographies is much faster.

PR #9, opened 2018-01-22
A filter for short captions in list of figures for latex output

gtuckerkellogg:
It uses a figure attribute 'short-caption'. I created this for my own students who have to write dissertations including lists of figures. 

The README file includes an example ( `make readme` will create an example pdf with documentation).

It works with `pandoc-crossref` regardless of the order of application.

Tested in pandoc 2.1.1


short-captions/short-captions.lua
```diff
@@ -0,0 +1,28 @@
...
+   return pandoc.RawInline('latex', str)
+end
+
+function Para (figure)
+   if (FORMAT =="latex") and (#figure.content == 1)
+   and (figure.content[1].t == 'Image')  and (figure.content[1].title == "fig:") then
+      local img = figure.content[1]
+      if img.caption and img.attributes['short-caption'] then
+         local caption_cmd = string.format('\n\\caption[%s]',img.attributes['short-caption'])
+         local name = img.c[1][1]
```
tarleb: `img.identifier` would be a bit easier to read.  I guess we haven't really documented the existence of that accessor yet.

short-captions/short-captions.lua
```diff
@@ -0,0 +1,28 @@
+
+local function latex(str)
+   return pandoc.RawInline('latex', str)
+end
+
+function Para (figure)
+   if (FORMAT =="latex") and (#figure.content == 1)
```
tarleb: The paragraph is always marshalled to Lua whenever `Para` is invoked. Making this invocation conditional by moving `(FORMAT == 'latex')` out of this function would make the filter almost free when converting to other formats.

gtuckerkellogg: Fixed

PR #10, opened 2018-01-23
bug fix for special case of figure called "fig:" 

gtuckerkellogg:
I realised there was a bug for the edge case in which someone creates a figure titled "fig:". Sorry about missing that earlier. 

Also now supports pandoc markdown inside the short caption


tarleb:
Thanks!

PR #12, opened 2018-01-23
Track changes in LaTeX or accept it in other output formats

tolot27:
This filters can be used to accept all tracked changes in any output format or to convert them to be visible in LaTeX/PDF.


track-changes/README.md
```diff
@@ -0,0 +1,14 @@
+# Tracks changes in LaTeX or removes them in other output formats
+
+The Pandoc Docx reader and writer supports track changes of MS Word (command line parameter `--track-changes=accept|reject|all`).
+
+According to the [manual](https://pandoc.org/MANUAL.html) the option `accept` is the default
```
tarleb: This could be a documentation bug in pandoc from the sound of it. Please raise an issue there if you think it is.

track-changes/track-changes-latex.lua
```diff
@@ -0,0 +1,102 @@
...
+\newcommand{\note}[2][]{\added[#1,remark={#2}]{}}
+\newcommand\hl{%
+  \bgroup
+  \expandafter\def\csname sout\space\endcsname{\bgroup \ULdepth =-.8ex \ULset}%
+  \markoverwith{\textcolor{yellow}{\rule[-.5ex]{.1pt}{2.5ex}}}%
+  \ULon}
+]]
+
+local function initials(s)
+    ignore = { -- list of words to ignore
```
tarleb: missing `local`?

track-changes/track-changes-latex.lua
```diff
@@ -0,0 +1,102 @@
...
+    return table.concat(ans)
+end
+
+local toTex = {["comment-start"] = "\\protect\\note", insertion = "\\added", deletion = "\\deleted"}
+
+function M.Span(elem)
+    if toTex[elem.classes[1]] ~= nil then
+        local author = elem.attributes.author
+        local inits
+        if author:find(" ") then inits = initials(author) else inits = author end
```
tarleb: a common idiom in lua is to use something like `inits = author:find' ' and initials(author) or author`, which is shorter and arguably(!) easier to read.

track-changes/track-changes-latex.lua
```diff
@@ -0,0 +1,102 @@
...
+end
+
+local function pairsByKeys(t, f)
+    local a = {}
+    for n in pairs(t) do table.insert(a, n) end
+    table.sort(a, f)
+    local i = 0      -- iterator variable
+    local iter = function ()   -- iterator function
+      i = i + 1
+      if a[i] == nil then return nil
```
tarleb: Could this test be omitted? Both `a[i]` and `t[a[i]]` should be `nil` when `i > #a`.

track-changes/track-changes-latex.lua
```diff
@@ -0,0 +1,102 @@
+
+local List = require 'pandoc.List'
```
tarleb: `List` appears to be unused.

track-changes/track-changes-final.lua
```diff
@@ -0,0 +1,9 @@
+function Span(elem)
```
tarleb: Starting with pandoc 2.1.1, the value of `--track-changes` is available in `PANDOC_READER_OPTIONS.trackChanges`. Would it make sense to use this option to alter the behavior of this filter?

track-changes/README.md
tolot27: Okay, makes sense and I've already thought about it. Now it is raised: https://github.com/jgm/pandoc/issues/4301

track-changes/track-changes-latex.lua
tolot27: Yea, a copy left-over from https://github.com/tolot27/lua-filters/blob/master/task-list/task-list.lua#L1

track-changes/track-changes-latex.lua
tolot27: Since I'm new to lua, I've looked for a specification of ternary operators but could only find discussions about it - also with mentioned drawbacks. That's why I used the "full" version which was easier to read for me. Now, I changed them as suggested.

track-changes/track-changes-latex.lua
tolot27: I got this from here: https://www.lua.org/pil/19.3.html. In principle, the sorting of the authors could be removed. But then `make` will fail sometimes because the output differs from the `expected_draft.tex` file. I don't know how to avoid this.

tolot27: Interestingly, just the `return a[i], t[a[i]]` works, even if I currently don't understand it. Is this safe?

track-changes/track-changes-final.lua
tolot27: Great to know this. I wanted to ask this soon on the discussion group.

tolot27:
I'll provide a sample docx as soon as jgm/pandoc#4303 is fixed.

tolot27:
Because now I use PANDOC_READER_OPTIONS.trackChanges, available with Pandoc version 2.1.1, Travis builds fail. How to solve this problem?

tarleb:
You could define a default behavior, or allow users to set behavior via a metadata field, if reader options are not available:

``` lua
-- this is untested
local track_changes = PANDOC_READER_OPTIONS
    and PANDOC_READER_OPTIONS.trackChanges
    or meta['trackChanges'] or 'AcceptChanges'
```

tolot27:
Now, meta is not available. I tried to query it in `function Pandoc(doc)` but it looks like this function get executed after `return M`. Any idea how to access it earlier or assign the functions in `M[1]` in another way?

tarleb:
Thanks for taking the time to get things right. Is this ready to merge?

tolot27:
Yes, it can be merged. There are still some issues open in pandoc (jgm/pandoc/issues/4427, jgm/pandoc/issues/4305, jgm/pandoc/issues/4456, jgm/pandoc/issues/4609] which prevents roundtripping of the sample docx. But the main functionality works very well.

tolot27:
Are there any further concerns or suggestions preventing the merge of this PR?

tarleb:
Apologies. I meant to merge this a long time ago, but forgot.

PR #13, opened 2018-01-30
Update Readme of short-captions

geniusupgrader:
I've added that it works with pandoc-crossref and that it does not work on tables and listings yet

### DIFF
--- a/short-captions/README.md
+++ b/short-captions/README.md
@@ -5,7 +5,8 @@ lof: true
 
 # Short captions in \LaTeX\ output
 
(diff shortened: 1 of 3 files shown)


tarleb:
Thank you. I wonder whether it would be more appropriately to place this in the section on *limitations*.

geniusupgrader:
@tarleb I have changed it to the limitations sections. Will you merge it?

tarleb:
Thanks. Could you take a look at the test-failure? The tests for this filter use the README as input; any changes must be reflected in the expected output.

PR #16, opened 2018-01-30
update license year

svenevs:
Thanks for the filter examples, I'm really looking forward to getting rid of my ridiculous post-processing and writing lua filters :smile:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2017 pandoc
+Copyright (c) 2017-2018 pandoc
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


tarleb:
Thanks. Could change this into a range of years? Citing the GNU project's [advice on copyright notices](https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html):

> Don’t delete old year numbers, though; they are significant since they indicate when older versions might theoretically go into the public domain, if the movie companies don’t continue buying laws to further extend copyright. If you copy a file into the package from some other program, keep the copyright years that come with the file.

svenevs:
Righto, thanks for the link, I'm going to read the article more carefully now!

I rebased with master and force-pushed a different commit to keep the history a little cleaner.

PR #19, opened 2018-03-03
fix #18: overwrite of header-includes

### LINKED ISSUE #18: TaskList filter conflict with YAML metadata
The `task-list.lua` filter currently eats up any `header-includes` defined via in YAML headers.

I'm invoking pandoc `v2.1.1` via the following batch script:

```batch
pandoc --no-highlight ^
       --from       markdown_github+smart+yaml_metadata_block+auto_identifiers ^
       --to         html5 ^
       --template   ./architect.template ^
       --output     ./index.html ^
       --lua-filter=task-list.lua ^
       ../README.md ./configuration.yaml
```

where the `./configuration.yaml` file contains some `header-includes` entries:

```yaml
header-includes:
  - |
    <!-- anything you put here will go verbatim in the header section
         Useful for adding javascripts, custom metadata, and so on... -->
  - |
    <!-- another -->
```

these entries don't show up in the final HTML doc; but if I remove the `--lua-filter=task-list.lua` option from the script, they show up as intended.

I've looked at the source of the Task List filter, and it looks like it does check if elements are present in the list of  `header-includes`, but apparently it fails when YAML headers are involved:

```lua
--- Add task-list CSS styles to the header.
function M.add_task_list_css(meta)
  local header_includes
  if meta['header-includes'] and meta['header-includes'].t ~= 'MetaList' then
    header_includes = meta['header-includes']
  else
    header_includes = pandoc.MetaList{meta.header_includes}
  end
  header_includes[#header_includes + 1] =
    pandoc.MetaBlocks{pandoc.RawBlock('html', M.css_styles)}
  meta['header-includes'] = header_includes
  return meta
end
```

tolot27:


tarleb:
Thanks!

PR #28, opened 2018-05-10
Add filter to move abstract from body to metadata

tarleb:
This moves a document's abstract from the main text into the
metadata. Metadata elements usually allow for finer placement
control in the final output, but writing body text is easier and
more natural.


jgm:
PS. I don't know if you know that you have write access to this repository.

tarleb:
Yes, sorry. Old habit. I'll just push stuff like this from now on.

PR #36, opened 2018-12-30
other-language: 

greut:
A filter I've been using to avoid french hyphenation in code blocks once transformed into a TeX document.

**source**

```markdown
```yaml
title: Mon titre
lang: fr
```

**without**

screenshot from 2018-12-30 15-05-10

**with**

screenshot from 2018-12-30 15-03-59

Extracted from this project: https://github.com/HE-Arc/rapport-technique

### DIFF
--- a/other-language/Makefile
+++ b/other-language/Makefile
@@ -1,0 +1,6 @@
+test:
+	@pandoc --lua-filter=other-language.lua --output=output.tex sample.md
+	@diff -u expected.tex output.tex
+	@rm -f output.tex
+
+.PHONY: test
--- a/other-language/README.md
+++ b/other-language/README.md
@@ -1,0 +1,37 @@
+# other-language for TeX code blocks
+
+This filter ensures that the raw content such a code blocks are enclosed with
+the english language so any typographic rules existing the base language are
+not applied. It was conceived with French in mind but surely can be useful to
+other languages.
+
+## Using the filter
+
+For it to do anything, the main document language should be something else than
+`en`, e.g. `fr`.
+
+```yaml
+---
+lang: fr
+---
+```
+
+Which will produce
+
+```latex
+\documentclass[french,]{article}
+
+% ...
+
+\usepackage[shorthands=off,english,main=french]{babel}
+
+% ...
+
+\usepackage{polyglossia}
+\setmainlanguage[]{french}
+\setotherlanguage[]{english}
+```
+
+Then any code block, or inline code will be surrounded with the
+`\begin{otherlanguage}{english}` and `\end{otherlanguage}`. It will be killing
+off any typographic changes such as the insecable space before a colon.
--- a/other-language/expected.tex
+++ b/other-language/expected.tex
@@ -1,0 +1,20 @@
+\begin{otherlanguage}{english}
+
+\begin{Shaded}
+\begin{Highlighting}[]
+\FunctionTok{lang:}\AttributeTok{ fr}
+\end{Highlighting}
+\end{Shaded}
+
+\end{otherlanguage}
+
+\begin{otherlanguage}{english}
+
+\begin{verbatim}
+lang: fr
+\end{verbatim}
+
+\end{otherlanguage}
+
+In a paragraph,
+\begin{otherlanguage}{english}\texttt{lang:\ fr}\end{otherlanguage}.
--- a/other-language/other-language.lua
+++ b/other-language/other-language.lua
@@ -1,0 +1,43 @@
+local lang = 'english'
+
+local before = pandoc.RawInline(
+  'tex', '\\begin{otherlanguage}{' .. lang .. '}')
+local after = pandoc.RawInline(
+  'tex', '\\end{otherlanguage}')
+
+local meta = {}
+
+return {
+  {
+    Meta = function (el)
+      meta = el
+      meta['babel-otherlangs'] = lang
+      meta['polyglossia-otherlangs'] = {name = lang}
+      return {}
+    end,
+  }, {
+    Code = function(code)
+      return {before, code, after}
+    end,
+
+    CodeBlock = function(code_block)
+      return {
+        pandoc.Para({before}),
+        code_block,
+        pandoc.Para({after})
+      }
+    end,
+
+    RawBlock = function(raw_block)
+      return {
+        pandoc.Para({before}),
+        raw_block,
+        pandoc.Para({after})
+      }
+    end,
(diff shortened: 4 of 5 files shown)


jgm:
It's surprising to me that polyglossia does these transformations inside verbatim contexts.
(I assume it's polyglossia that is doing it?)

Is it because pandoc's highlighted code blocks use a custom environment that polyglossia doesn't recognize?  If you just have a code block without syntax marked, does it do the same thing?

Just wondering whether this should be handled with a modification to pandoc rather than a filter.

greut:
It seems to do the same thing.

        ---
        title: Mon titre
        lang: fr
        ---

    ```
    ---
    title: Mon titre
    lang: fr
    ---
    ```

    ```yaml
    ---
    title: Mon titre
    lang: fr
    ---
    ```

screenshot from 2019-01-01 11-51-59

jgm:
Interestingly, this only happens with `--pdf-engine=xelatex`.

jgm:
From polyglossia manual:

> In some very specific contexts (such as music score creation), TEX hyphenation is something to avoid as it may cause troubles. polyglossia provides two functions: `\disablehyphenation` and `\enablehyphenation`. Note that when you se- lect a new language, hyphenation will be in the same state (enabled or disabled) as before. When you reenable it, it will take the last selected language.

jgm:
I don't know if it's worth adding all the additional complexity to the default template that this would require to redefine the verbatim environments.  But for this filter, it would probably be cleaner to add `\disablehyphenation` before each code block and `\enablehyphenation` after it, instead of using `otherlanguage`.  (And then it would make sense to call the filter something like `disable-hyphenation-in-code`.)

greut:
that would have been so neat and easy, but I didn't manage to get it working.

screenshot from 2019-01-01 22-01-53


```tex
\documentclass[]{article}

\usepackage{polyglossia}
\setmainlanguage[]{french}
\setotherlanguage[]{english}

\begin{document}

lang: fr

\disablehyphenation
lang: fr-nohyphen
\enablehyphenation

\begin{otherlanguage}{english}
lang: en
\end{otherlanguage}

\end{document}
```

jgm:
I guess that makes sense:  it's not really a matter of hyphenation, after all.
I looked through the polyglossia manual, and it doesn't say anything about colons.  So, maybe your approach is the only one that will work!

jgm:
But I do think the name of this filter is confusing.  Wouldn't a better name somehow mention code or verbatim blocks, so people will know what this pertains to?

jgm:
PS. Instead of adding a Para before and after containing raw latex, you might try putting each code block inside a Div with attribute lang="en".  This works just as well and it's cleaner and simpler.

greut:
I need a little help with the lua side though... sorry.

When putting the `<div lang="en">` in the markdown, it produces this.

    Div ("",[],[("lang","en")]) [ ... ]

Although, in the following Lua `pandoc.Div(el, pandoc.Attr("lang", {"en"}))` produces.

    Div ("lang",["en"],[]) [ ... ]

Thanks!

greut:
Huzzah!

screenshot from 2019-01-02 20-12-11

I'll reopen a new PR soon. Thanks a lot!

adunning:
This does appear to be a bug specific to Polyglossia and XeTeX. It works fine with Babel:

```latex
\documentclass{article}

\usepackage{color}
\usepackage{fancyvrb}
\newcommand{\VerbBar}{|}
\newcommand{\VERB}{\Verb[commandchars=\\\{\}]}
\DefineVerbatimEnvironment{Highlighting}{Verbatim}{commandchars=\\\{\}}
% Add ',fontsize=\small' for more characters per line
\newenvironment{Shaded}{}{}
\newcommand{\AttributeTok}[1]{\textcolor[rgb]{0.49,0.56,0.16}{#1}}
\newcommand{\FunctionTok}[1]{\textcolor[rgb]{0.02,0.16,0.49}{#1}}
\newcommand{\OtherTok}[1]{\textcolor[rgb]{0.00,0.44,0.13}{#1}}

\usepackage[shorthands=off,main=french]{babel}

\begin{document}

\begin{Shaded}
\begin{Highlighting}[]
\OtherTok{---}
\FunctionTok{title:}\AttributeTok{ Mon titre}
\FunctionTok{lang:}\AttributeTok{ fr}
\OtherTok{---}
\end{Highlighting}
\end{Shaded}

\end{document}
```

adunning:
There's a workaround for Polyglossia posted in <https://github.com/reutenauer/polyglossia/issues/27>.

jgm:
Good find. We could simply add this to the default
pandoc template where polyglossia is loaded:

    \makeatletter
    \appto\verbatim@font{\nofrench@punctuation}
    \makeatother

It's a bit verbose to have all of this in the
template, since it's irrelevant unless lang is fr.
I suppose we could have the writer check the
lang in metadata, and set a special variable if
lang = fr; these lines could be included if the
variable is set.

If this seems a good idea, perhaps an issue should
be opened for pandoc.  (Though if we decide to make
polyglossia optional, not default, for xelatex,
then it may be a nonissue.)

jgm:
@greut - given the workaround, is this filter even necessary any more?

greut:
@jgm I cannot get the work around to work. Although, it became clear that polyglossia+xelatex are the bad guys here, I'll keep this filter and consider it a hack rather than something worth spreading. Thanks everyone for you help!

adunning:
@greut Are you receiving an error, or is it simply not displaying properly? This works for me under either XeTeX or LuaTeX:

```latex
\documentclass{article}

% Polyglossia with workaround
\usepackage{polyglossia}
\setdefaultlanguage{french}
\setotherlanguage{english}

\makeatletter
\appto\verbatim@font{\nofrench@punctuation}
\makeatother
% 

% or use Babel
% \usepackage[english, main=french]{babel}

\begin{document}

\verb|00:11:43:D4:86:A0|

\selectlanguage{english}
\verb|00:11:43:D4:86:A0|

\selectlanguage{french}
\begin{verbatim}
00:11:43:D4:86:A0
\end{verbatim}

\end{document}
```

Note that you might have to delete your .aux file between runs when switching between Babel and Polyglossia. If you're still having trouble, perhaps try checking that you're running the latest version of TeX Live by running `tlmgr update --self --all --reinstall-forcibly-removed`.

greut:
pandoc produces `\texttt{}` and uses `fancyvrb` which are unaffected by the fix. No worries.

PR #40, opened 2019-01-10
Minted filter for beamer and latex, safe for all writers

svenevs:
Hey!  I made a new years resolution to get this all nice and shiny **and** in `lua`, and I think it's looking pretty good so far.

- [x] initial filter and documentation
- [x] finish addressing issues / suggestions in tarleb's review!

### DIFF
--- a/minted/minted.lua
+++ b/minted/minted.lua
@@ -406,8 +406,6 @@ function Header(elem)
     -- may duplicate the 'fragile' specifier.  Duplicated fragile does *not*
     -- cause compile errors.
     return elem
-  else
-    return elem
   end
 end
 
--- a/minted/minted.lua
+++ b/minted/minted.lua
@@ -193,12 +193,10 @@ local function minted_attributes(elem, type)
 
   -- Boolean style options for minted (e.g., ```{.bash .autogobble}) will appear
   -- in the list of classes.
-  if #elem.classes > 0 then
-    for _, cls in ipairs(elem.classes) do
-      if is_minted_class(cls) then
-        table.insert(minted_attributes, cls)
-        table.insert(minted_keys, cls)
-      end
+  for _, cls in ipairs(elem.classes) do
+    if is_minted_class(cls) then
+      table.insert(minted_attributes, cls)
+      table.insert(minted_keys, cls)
     end
   end
 
--- a/minted/minted.lua
+++ b/minted/minted.lua
@@ -281,15 +281,17 @@ function Meta(m)
     end
 
     -- Parse and set the default block language.
-    local default_block_language = minted["default_block_language"]
+    local default_block_language = minted.default_block_language
+      and pandoc.utils.stringify(minted.default_block_language)
     if default_block_language ~= nil then
-      minted_default_block_language = default_block_language[1].text
+      minted_default_block_language = default_block_language
     end
 
     -- Parse and set the default inline language.
-    local default_inline_language = minted["default_inline_language"]
+    local default_inline_language = minted.default_inline_language
+      and pandoc.utils.stringify(minted.default_inline_language)
     if default_inline_language ~= nil then
-      minted_default_inline_language = default_inline_language[1].text
+      minted_default_inline_language = default_inline_language
     end
 
     -- Parse the global default minted attributes to use on every block.
--- a/minted/minted.lua
+++ b/minted/minted.lua
@@ -1,0 +1,430 @@
+--[[
+minted -- enable the minted environment for code listings in beamer and latex.
+
+MIT License
+
+Copyright (c) 2019 Stephen McDowell
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+]]
+
+--------------------------------------------------------------------------------
+-- Quick documentation.  See full documentation here:                         --
+-- https://github.com/pandoc/lua-filters/blob/master/minted                   --
+--------------------------------------------------------------------------------
+--[[
+Brief overview of metadata keys that you can use in your document:
+
+minted:
+  no_default_autogobble:   <boolean>, *DISCOURAGED*
+  no_mintinline:           <boolean>
+  default_block_language:  <string>
+  default_inline_language: <string>
+  block_attributes:        <list of strings>
+    - attr_1
+    - attr_2
+    - ...
+  inline_attributes:       <list of strings>
+    - attr_1
+    - attr_2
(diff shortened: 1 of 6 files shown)


minted/minted.lua
```diff
@@ -0,0 +1,396 @@
...
+--------------------------------------------------------------------------------
+-- Return the string lexer class to be used with minted.  `elem` should be
+-- either a Code or CodeBlock element (whose `classes` list will be inspected
+-- first).  `type` is assumed to be either `MintedInline` or `MintedBlock` in
+-- order to choose the appropriate fallback lexer when unspecified.
+local function minted_language(elem, type)
+  -- If the code [block] attached classes, we assume the first one is the
+  -- lexer class to use.
+  if #elem.classes > 0 then
+    return elem.classes[1]
```
svenevs: So right now, if somebody does `{.showspaces .bash}` instead of `{.bash .showspaces}` this will break.  I intend on making the documentation very explicit that the lexer class name must always be first, **and** must always be specified.  That is, they cannot just do `{.showspaces}`, they must include the lexer class.

Does this seem ok?  I think one of the other filters does this, but I don't really want to go through all of `pygmentize -L` and try and find the language "intelligently".  Any alternatives I'm not seeing?

minted/minted.lua
```diff
@@ -0,0 +1,396 @@
...
+      if is_minted_class(cls) then
+        table.insert(minted_attributes, cls)
+        table.insert(minted_keys, cls)
+      end
+    end
+  end
+
+  -- Value options using xxx=value (e.g., ```{.bash fontsize=\scriptsize}) show
+  -- up in the list of attributes.
+  for _, attr_table in ipairs(elem.attr[3]) do
```
svenevs: I found this by just printing things endlessly until I found them.  There's a distinction that `pandoc` seems to make, the rules are

1. Boolean flags must have a `.` in front, so `{.showspaces}` not `{showspaces}`
2. Things with an `=` sign cannot have a `.` in front, so `{bgcolor=red}` not `{.bgcolor=red}`

This part is for (2), where it seems like they reliably show up at the third element, and are also conveniently already split up for me.  That's why `cls, value = attr_table[1], attr_table[2]` happens, `pandoc` already splits into `bgcolor` and `red` for example.

The fact that they are already split up makes me think there is a more direct way to get this, I just don't know what it is.  Is there a better way?

minted/minted.lua
```diff
@@ -0,0 +1,396 @@
...
+  for _, cls in ipairs(elem.classes) do
+    if not is_minted_class(cls) then
+      table.insert(classes, cls)
+    end
+  end
+  elem.classes = classes
+
+  -- Remove any minted items from the attributes.
+  extra_attrs = {}
+  for _, attr_table in ipairs(elem.attr[3]) do
```
svenevs: Instance number two of magic `.attr[3]` call.

minted/minted.lua
```diff
@@ -0,0 +1,396 @@
...
+function Meta(m)
+  -- Grab the `minted` metadata, quit early if not present.
+  local minted = m["minted"]
+  local found_autogobble = false
+  local always_autogobble = true
+  if minted ~= nil then
+    -- Parse and set the default block language.
+    local default_block_language = minted["default_block_language"]
+    if default_block_language ~= nil then
+      minted_default_block_language = default_block_language[1].text
```
svenevs: Similarly, is there a more correct way to get `default_block_language` text out without this awkward (and also found by accident) `[1]` here?  I have no idea if this is valid or not, I was just printing out types of things and looping through tables until I found what I knew it was supposed to be .... .... :upside_down_face:

minted/minted.lua
```diff
@@ -0,0 +1,396 @@
...
+    -- NOTE: pass the remaining work to pandoc, noting that 2.5 and below
+    -- may duplicate the 'fragile' specifier.  Duplicated fragile does *not*
+    -- cause compile errors.
+    return el
+  else
+    return elem
+  end
+end
+
+-- NOTE: order of return matters, Meta needs to be first otherwise the metadata
```
svenevs: This is more just curious to me.  If I don't do a `return` here at all (since the methods are named as they are `pandoc` will still use them), putting `print` statements in you will see that `CodeBlock` runs before `Meta`.  Is this a valid solution to this, or is there a more correct way to ensure that I can parse the `Meta` first and populate the "filter level variables"?

jgm:
Stephen McDowell <notifications@github.com> writes:

> +    -- NOTE: using "latex" here because using FORMAT as a parameter does
> +    -- not seem to work.  It should not cause problems for beamer.
> +    return pandoc.RawBlock("latex", raw_minted)
>
> Sorry, this one occurs in two places and I missed both in the last review.  Why can't I do `return pandoc.RawBlock(FORMAT, raw_minted)`?  If I do that then nothing shows up (aka the code blocks just disappear).

That's a bit of a rough edge in pandoc.  "beamer"
doesn't currently work as a raw format.  I've just
pushed a commit fixing that, but for now use "latex".

tarleb:
Great! Feel free to squash,  but we can also squash this when merging (we generally do this for all PRs introducing new filters).

svenevs:
ok i think this looks good :slightly_smiling_face:

i hope the tests being written in python is ok, i had to do a lot of string manipulation and that's definitely my go-to language for that.  grepping for the latex was annoying because latex syntax definitely conflicts with regex syntax.

sorry for multiple force pushes :D but now it's just one commit, unless things need to change in which case future commits will be structured better and can just be squash merged on your end.

minted/minted.lua
```diff
@@ -0,0 +1,430 @@
...
+    if not has_fragile then
+      table.insert(elem.classes, 'fragile')
+    end
+
+    -- NOTE: pass the remaining work to pandoc, noting that 2.5 and below
+    -- may duplicate the 'fragile' specifier.  Duplicated fragile does *not*
+    -- cause compile errors.
+    return elem
+  else
+    return elem
```
tarleb: Not returning anything (or explicitly returning `nil`) will leave the element unchanged and is a bit faster, as the element doesn't need to be demarshaled. But the difference in performance is minor, and being explicit about what's returned has its merits, too.

minted/minted.lua
tarleb: I agree that this behavior is unfortunate. It's a known issue and on my todo list.

minted/minted.lua
svenevs: HECK YES, `.attributes` is *way* better than `.attr[3]` :slightly_smiling_face:

minted/minted.lua
svenevs: I don't think it's that hidden, I got the `.classes` right from those docs...I just didn't connect the dots somehow that I could `.attributes` :scream:

minted/minted.lua
```diff
@@ -0,0 +1,430 @@
...
+    "linenos", "numberfirstline", "numbers", "mathescape", "numberblanklines",
+    "numbersep", "obeytabs", "outencoding", "python3", "resetmargins",
+    "rulecolor", "samepage", "showspaces", "showtabs", "space", "spacecolor",
+    "startinline", "style", "stepnumber", "stepnumberfromfirst",
+    "stepnumberoffsetvalues", "stripall", "stripnl", "tab", "tabcolor",
+    "tabsize", "texcl", "texcomments", "xleftmargin", "xrightmargin"
+  }
+
+  -- Try and find the specified `cls` in `all_minted_options`.
+  for _, minted_option in ipairs(all_minted_options) do
```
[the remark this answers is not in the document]
svenevs: perfect!

minted/minted.lua
svenevs: yes, this is much cleaner!

svenevs:
Ok round two complete, thanks for the really helpful review on making this much cleaner @tarleb!  All of your feedback has been incorporated and a few updates to the tests to actually validate the `Header` (`ensure_not_present(test_name, "fragile", output)`) in the non-beamer stuff, and make sure metadata doesn't produce `nil`.  I think this looks really good now :slightly_smiling_face: 

IIRC when squashed the title of the PR is what the main commit message is.  I don't know what the main message should be, but feel free to edit it to be whatever you want.

tarleb:
Thanks again! This is probably the best documented and tested filter in the collection.

svenevs:
Woohoo, you're very welcome!  Hehe yes ... I got a little carried a way xD

mfhepp:
Am impressed by how well this filter is written and documented! Dedication to detail at its best :-)! 

I hope this is the best place to leave this info for others: For the ArXiv preprints workflow, there are some limitations on using minted. In essence, it is important to use a non-hidden cache directory, like so

`\usepackage[frozencache=true,cachedir=minted-cache]{minted} `

The important thing is that the `cachedir` name does not start with a dot as the default `.minted-cache`.

Full details are here: https://arxiv.org/help/faq/mistakes#minted

Disclaimer: I did not test this yet, but thought it is worth logging it for the future.

PR #43, opened 2019-01-20
minted: Fix inline code when it contains special characters.

### LINKED ISSUE #42: minted: Bug in mintinline with unbalanced braces
Consider the following text:
```markdown
C uses `{` and `}` to delimit scopes.
```
which converts to the following LaTeX:
```latex
C uses \mintinline[]{text}{{} and \mintinline[]{text}{}} to delimit scopes.
```

For `\mintinline`, if using braces for the final parameter, it extends to balanced braces. The resulting PDF looks something like:

> C uses `{} and \mintinline[]{text}{}` to delimit scopes.

For these cases, `\mintinline` works like `\verb` in that it can accept almost any other character as a delimiter, e.g., `\mintinline[]{text}|{| and \mintinline[]{text}|}|` would work in this case. I opened a PR to the Python version of this filter nick-ulle/pandoc-minted#8 just before I found this one. Unfortunately, I'm not as familiar with Lua so I can't do the same here.

cc @svenevs

QuLogic:
If an inline code section contains an unmatched brace, then using braces as the `\mintinline` delimiter breaks. The inline section will extend to the next matching brace instead.

`\mintinline` can use any other character for a delimiter, so try to switch to something else when braces are in the text.

Fixes #42.

### DIFF
--- a/minted/minted.lua
+++ b/minted/minted.lua
@@ -349,13 +349,37 @@ function Code(elem)
       return pandoc.RawInline("latex", string.format("\\texttt{%s}", elem.text))
     end
 
+    local start_delim, end_delim
+    if elem.text:find('[{}]') then
+      -- Try some other delimiter (the alphanumeric digits are in Python's
+      -- string.digits + string.ascii_letters order)
+      local possible_delims = ('|!@#^&*-=+' .. '0123456789' ..
+                               'abcdefghijklmnopqrstuvwxyz' ..
+                               'ABCDEFGHIJKLMNOPQRSTUVWXYZ')
+      for char in possible_delims:gmatch('.') do
+        if not elem.text:find(char, 1, true) then
+          start_delim = char
+          end_delim = char
+          break
+        end
+      end
+      if not start_delim then
+        local msg = 'Unable to determine delimiter to use around inline code %q'
+        error(msg:format(elem.text))
+      end
+    else
+      start_delim = '{'
+      end_delim = '}'
+    end
     local language   = minted_language(elem, MintedInline)
     local attributes = minted_attributes(elem, MintedInline)
     local raw_minted = string.format(
-      "\\mintinline[%s]{%s}{%s}",
+      "\\mintinline[%s]{%s}%s%s%s",
       attributes,
       language,
-      elem.text
+      start_delim,
+      elem.text,
+      end_delim
     )
     -- NOTE: prior to pandoc commit 24a0d61, `beamer` cannot be used as the
     -- RawBlock format.  Using `latex` should not cause any problems.
--- a/minted/minted.lua
+++ b/minted/minted.lua
@@ -262,6 +262,38 @@ local function remove_minted_attibutes(elem)
   return elem
 end
 
+-- Return a `start_delim` and `end_delim` that can safely wrap around the
+-- specified `text` when used inline. If no special characters occur in `text`,
+-- then a pair of braces are returned. Otherwise, if any character of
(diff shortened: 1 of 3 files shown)


minted/run_minted_tests.py
```diff
@@ -313,17 +328,16 @@ def verify(test_name, md, string):
...
+        r"\texttt{{}",
+        r"\texttt{}}"
     )
     verify(
         "[inline-code] .no_minted class bypasses single inline code element",
         inline_code.replace("{.cpp}", "{.cpp .no_minted}"),
-        r"\texttt{#include <type_traits>}"
-    )
-    verify(
-        "[inline-code] default inline language is 'text'",
```
QuLogic: Note: this test is the same as the one on line 303/316.

minted/run_minted_tests.py
```diff
@@ -333,7 +347,7 @@ def verify(test_name, md, string):
               default_inline_language: "haskell"
             ---
             {inline_code}
-        ''').format(inline_code=inline_code.replace("{.cpp}", "")),
+        ''').format(inline_code=inline_code),
```
QuLogic: I removed the `replace` since the plain `{` will test this.

minted/sample.md
```diff
@@ -30,28 +30,30 @@ minted:
 - Raw inline code:
 
     ```md
-    `#include <type_traits>`
+    `#include <type_traits>` `{ int x; }`
```
QuLogic: I chose not to put the whole set of delimiters into the sample, but I did try out once with _all_ delimiters but one, and it worked reasonably.

QuLogic:
Oops, the code I put into the `sample.md` doesn't demonstrate the bug; fixed.

minted/minted.lua
```diff
@@ -346,16 +346,53 @@ function Code(elem)
 
     -- Check for local or global bypass to turn off \mintinline
     if minted_no_mintinline or found_no_minted_class then
-      return pandoc.RawInline("latex", string.format("\\texttt{%s}", elem.text))
+      local text = elem.text:gsub(
```
tarleb: What would happen if we were to just return `elem` unchanged here? Would that still work?

QuLogic: Oh, does that let it fall back to pandoc's default `CodeInline` implementation? That would certainly be simpler.

QuLogic: It does appear to do so, so I've changed it to work this way.

minted/minted.lua
```diff
@@ -346,16 +346,40 @@ function Code(elem)
 
     -- Check for local or global bypass to turn off \mintinline
     if minted_no_mintinline or found_no_minted_class then
-      return pandoc.RawInline("latex", string.format("\\texttt{%s}", elem.text))
+      return elem
     end
 
+    local start_delim, end_delim
```
tarleb: Putting all this into a separate function might improve readability.

QuLogic: Done.

svenevs:
@QuLogic this is **awesome** -- you're fixes here are great!  Sorry for not seeing this until now.

I added a follow-up PR #44 to clarify the impact of leaving `Code` elements in tact rather than returning `\texttt` commands.  I love the improvements you made to this filter, and to be extra clear returning the unchanged element back to `pandoc` is a great decision -- the filter doesn't need to know how / what to escape for special characters :heart:  That PR is just documenting / testing the impact of `--no-highlight`.

QuLogic:
Thanks!

Though actually that bit was @tarleb's idea.

PR #44, opened 2019-01-29
document/test impact of `--no-highlight` on minted filter

svenevs:
- [x] Add `--no-highlight` to `Makefile`
- [x] Document in README impact of `--no-highlight` flag and how this filter behaves.
- [x] Add explicit test in the sample to show how `--no-highlight` affects output.  (Deletions in that section are just to make it all fit in a single beamer slide).
- [x] Restructure latex tests to require pandoc arguments passed in as second parameter to `verify` local function.  Makes it so that default tests use `--no-highlight`, but two tests remove this and verify that the `pandoc` highlighting commands get used.

Only filter change is to `return nil` instead of `return elem` because @tarleb said that was slightly faster for `pandoc`.  Everything else is just documentation and testing impact of `--no-highlight` flag.

### DIFF
--- a/minted/Makefile
+++ b/minted/Makefile
@@ -23,14 +23,14 @@ pandoc_inputs := sample.md
 
 # Sample beamer presentation.
 sample_beamer.tex: $(pandoc_inputs)
-	pandoc -s -t beamer --lua-filter=minted.lua $^ > $@
+	pandoc -s -t beamer --no-highlight --lua-filter=minted.lua $^ > $@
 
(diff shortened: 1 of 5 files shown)


tarleb:
Cool! Good documentation like this always make me happy.

We plan to add access to writer options at some point in the future. If we do, then there could be a warning if `--no-highlight` has not been set. (jgm/pandoc#5221)

svenevs:
Ok cool just subscribed to that issue so that if it gets done I can add it here!

PR #45, opened 2019-02-11
Short-captions for Tables

blake-riley:
Tables in Markdown do not come with attributes (under the current AST), and there is no way to alter the `\caption{}` command written by Pandoc for tables in LaTeX.

- I was able to modify the behaviour of the caption command within a longtable, so that we can use the functionality of `\caption[short]{long}`. This was done by replacing one of the `\longtable` internal commands: `\LT@c@ption`.
This longtable macro now checks for a macro `\pandoctableshortcapt` when expanding the caption. It will use this custom macro in place of whatever is in the square brackets in `\caption[short]{long}`, even if there were no square brackets.
- After putting that macro overwrite in the header, the filter parses the caption of a Table.
  - If it finds a `short-caption` attribute, it places RawBlocks around the Table to define and undefine `\pandoctableshortcapt` appropriately.
  - If it finds a `.unlisted` class, it places RawBlocks around the Table to define `\pandoctableshortcapt` as `{}`. This causes the `\LT@c@ption` macro to not write an entry to the lof.

I've done a little bit of testing, and it seems to work nicely.

I have no idea how robust it is though, particularly to other changes in the header-includes.

### DIFF
--- a/table-short-captions/table-short-captions.lua
+++ b/table-short-captions/table-short-captions.lua
@@ -86,7 +86,8 @@ local function parse_table_attrs(attr)
   -- If not unlisted, then find the property short-caption.
   local short_caption = nil
   if not unlisted then
-    if attr.attributes["short-caption"] and (#attr.attributes["short-caption"] > 0) then
+    if (attr.attributes["short-caption"]) and 
+       (#attr.attributes["short-caption"] > 0) then
       short_caption = pandoc.read(attr.attributes['short-caption']).blocks[1].c
     end
   end
--- a/table-short-captions/table-short-captions.lua
+++ b/table-short-captions/table-short-captions.lua
@@ -73,23 +73,22 @@ end
 --- The undef shortcaption block to be placed after the table
 local undefshortcapt = pandoc.RawBlock('tex', "\\undef\\pandoctableshortcapt")
 
---- Search through a messy list of pandoc-parsed Inlines into what we need to build short-caption
--- @param (List of Inlines): A mess of Inlines containing the table properties
--- @return[1] (nil or string): The first word starting with "#tbl:"
+--- Treats a Span's Attr as Table Attr, and extracts what is needed to build short-caption
+-- @param (Attr): The Attr of the property Span in the table caption
+-- @return[1] (nil or string): The identifier
 -- @return[2] (nil or List of Inlines): The "short-caption" property if present, as a List of Inlines.
--- @return[3] (bool): Whether ".unlisted" appeared in the properties
-function parse_table_properties(props)
-  -- Flatten the string, because this will make our job much easier (except for short-caption hunting)
-  local flatprops = pandoc.utils.stringify(props)
-
+-- @return[3] (bool): Whether ".unlisted" appeared in the classes
+function parse_table_attrs(attr)
   -- Find label
-  local label = flatprops:match("(#tbl:%g-)%s")
+  local label = nil
+  if attr.identifier and (#attr.identifier > 0) then
+    label = attr.identifier
+  end
 
-  -- Find classes, particularly look for ".unlisted"
+  -- Look for ".unlisted" in classes
   local unlisted = false
-  classes = flatprops:gmatch("(%.[%w%-]-)[%s}]")
-  for c in classes do
-    if c == ".unlisted" then
+  for _, c in ipairs(attr.classes) do
+    if c == "unlisted" then
       unlisted = true
     end
   end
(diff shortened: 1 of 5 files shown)


blake-riley:
This PR is WIP because I haven't written a Makefile with tests yet.

blake-riley:
OK, tests done.

If you think it would be better merged into short-captions, please let me know.

tarleb:
I know very little about pandoc-crossref and LaTeX longtables, but this looks like an excellent idea.

One thing I'm not sure about: parsing the caption string is potentially fragile, and it would be nice if pandoc could handle for us. How about using a special Span to add this information? E.g.,

    Table: This is the *italicised long caption* of my table, which has a
    very long caption.
    []{#tbl:full-of-juicy-data short-caption="Short caption for *juicy* data table."}

It's two more characters, but one could just check for an empty span within the caption, and then transfer its attributes to the table.

tarleb:
Ah, I see: pandoc-crossref expects `{#tbl:table}` markers. Maybe using Span just for the short caption would work:

    Table: [short caption goes here]{.short-caption} The long caption follows {#tbl:table1}

blake-riley:
### pandoc-crossref

Yeah, sorry that wasn't particularly clear. The `{#tbl:table1}` is expected in the caption by pandoc-crossref—and so this filter deliberately sends it back in there.

I use pandoc-crossref with pandoc more often than not (I quite like its `[@tbl:crossref]` syntax). While I'm not too sure on how it works internally with tables, its net effect is to turn a `Table []` into `Div (tbl:table1) [Table []]` (at least when converting to native?). When converting to LaTeX, it puts a `\hypertarget{}` right before the table (I believe this is so links go to the top of the table, not to the caption), and inserts `\label{#tbl:table1}` into the caption.

I don't really want to reimplement what pandoc-crossref does for tables, so I left it to build the Div and hypertarget.

### Spans

**lightbulb moment** YES. Your suggestion of an empty span is completely the way to go here. I've just done some quick testing—I was completely unaware of how Pandoc parsed spans (and didn't think of them). Properties and all! Writing the caption parser felt pretty hacky, this is gonna be so much cleaner!

One of the aims for me was to make the syntax for table short-captions as similar as possible to the figure short-captions in this repo. I still think this is a reasonable aim?

### Todo

I'll rewrite the parser to target the following syntax tonight (your original suggestion):
```
Table: This is the long caption []{#tbl:table1 short-caption="short-caption here" .unlisted}
```

This should be far less fragile than the current hacky parsing. I'll still put `{#tbl:table1}` back into the caption for pandoc-crossref.

table-short-captions/table-short-captions.lua
```diff
@@ -0,0 +1,160 @@
...
+  \ifx\@tempa\@empty\else                 % If @tempa is blank, no lot entry! Otherwise, @tempa becomes the lot title.
+     {\let\\\space
+     \addcontentsline{lot}{table}{\protect\numberline{\thetable}{\@tempa}}}%
+  \fi}
+}\makeatother
+% -- end:latex-table-short-captions --
+]]
+
+--- Creates a def shortcaption block to be placed before the table
+-- @param sc: nil or a List of RawInlines
```
tarleb: One could also use LDoc's `@tparam` and `@treturn` when specifying type information. https://stevedonovan.github.io/ldoc/manual/doc.md.html

However, the docs didn't come out the way I expected when I tried last time, so just using the basic `@param` might actually be better. 👍

table-short-captions/table-short-captions.lua
```diff
@@ -0,0 +1,160 @@
...
+-- @return[3] (bool): Whether ".unlisted" appeared in the classes
+function parse_table_attrs(attr)
+  -- Find label
+  local label = nil
+  if attr.identifier and (#attr.identifier > 0) then
+    label = attr.identifier
+  end
+
+  -- Look for ".unlisted" in classes
+  local unlisted = false
```
tarleb: [`List.includes`](https://pandoc.org/lua-filters.html#pandoc.List:includes) would save a few lines.

table-short-captions/table-short-captions.lua
```diff
@@ -0,0 +1,160 @@
...
+    result:extend {undefshortcapt}
+  end
+  return result
+end
+
+--- Inserts longtable_caption_mod into the header_includes
+-- @param meta (Meta): The document metadata
+-- @return (Meta): The document metadata, with latex function inserted in header_includes
+function add_longtable_caption_mod(meta)
+  local header_includes
```
tarleb: Lua's excuse for a ternary operator can be useful here:

``` lua
local header_includes = -- <test if it's a MetaList>
    and meta['header-includes']
    or pandoc.MetaList(…)
```

This use of `and` and `or` has it's downsides (it's not *quite* the usual ternary operator), but it's idiomatic Lua. Some people prefer not to use it, though. Go with whatever you find the most readable.

table-short-captions/table-short-captions.lua
blake-riley: This is my first time writing Lua, so I have no idea about the doc package landscape.
I was just trying to copy someone else's code, so that it looked reasonable.
I've now installed luarocks and ldoc. (y)

table-short-captions/table-short-captions.lua
blake-riley: I was just doing what [`task-list.lua`](https://github.com/pandoc/lua-filters/blob/master/task-list/task-list.lua) did.
I don't mind a ternary, and from a Python background, the `and` and `or` style is what I'm used to (I'm assuming that it behaves the same).

blake-riley:
OK, last commit moved the logic slightly.
Rest of the patches should have resolved your comments.

tarleb:
Great, thank you!

PR #47, opened 2019-03-03
Test with Circle CI

tarleb:
Other pandoc projects use Circle CI now, so this one should as well. The only thing left is activating the project on Circle – @jgm, could you set this up?

@svenevs: this uses the pandoc LaTeX docker image, and it works nicely!


tarleb:
Having instructions for CircleCI in parallel to those for Travis CI doesn't hurt, so I've pushed the config. Circle CI is enabled on my personal fork and works. We can make the switch, run both in parallel, or just preserve the status quo.

Issue #50, opened 2019-04-06
Output of track-changes filter cannot be compiled to PDF when track-changes=all


tarleb:
Trying to compile sample.md to PDF via `pandoc -M trackChanges:all --track-changes=all --lua-filter=track-changes.lua --pdf-engine=xelatex --output sample.pdf sample.md` leads to the following error: 

    Error producing PDF.
    ! LaTeX Error: Option clash for package todonotes.

    See the LaTeX manual or LaTeX Companion for explanation.
    Type  H <return>  for immediate help.
     ...

    l.61 \setlength

Compiling with `track-changes=reject` or `track-changes=accept` seems to work ok.

@tolot27, could you take a look?

tolot27:
Should be fixed with PR #51.

tarleb:
Not fully fixed:

```
Error producing PDF.
! Undefined control sequence.
\Changes@Remark #1#2->\todo
                            [color=Changes@Color#1!20]{\sffamily \textbf {#1...
l.128 A \note[id=MCW]{I agree!}
```

tolot27:
Please can you provide more information about your tex environment? Most likely, you are using an older version of the changes package, maybe <3.x.

tarleb:
Right, I forgot. The problem *is* fixed if we test using our LaTeX docker image, but not when we use our current Travis CI setup. I have the same problem when testing locally (MacTeX 2018.0417).

You can use the commit from this branch to verify a fix: https://github.com/tarleb/lua-filters/tree/test-track-changes-pdf. This will allow us to test directly using Travis CI.

tolot27:
My newest compatibility patches should fix your last error message. Unfortunately, it does not work with pfdlatex (without `--pdf-engine=xelatex|lualatex`). The following error is reported (on different machines):

```
Error producing PDF.
This is pdfTeX, Version 3.14159265-2.6-1.40.20 (MiKTeX 2.9.7000)
(Fatal format file error; I'm stymied)
```

I'll have to look deeper.

tolot27:
Okay, got it to work. Please can you review it?

PR #51, opened 2019-04-08
fix for #50 caused by latex package changes v3.x

tolot27:
The latex package changes v3.x now loads the todonotes package by itself. That caused the option clash. This new version of the lua filter now requires changes v3.x.


tarleb:
Terrific, thanks! The change broke a test, it needs to be updated. I'm going to cherry-pick your commit manually for that.

PR #52, opened 2019-04-09
changes v2 and v3

tolot27:
fix compatibility and tests


tolot27:
This should fix the incompatibilities of #51 and solve #50 for more environments.

tarleb:
Looks good. Thank you for taking the time to fix this!

PR #57, opened 2019-05-03
Added pandoc-quotes.lua.

odkr:
`pandoc-quotes.lua` replaces non-typographic quotation marks with typographic ones for languages other than US English.


tarleb:
Thanks, this looks great and quite useful.

CI checks fail because access to the repo via git fails. While I like the idea of having a separate repo, it's worth noting that submodules have a couple of [disadvantages](https://www.atlassian.com/blog/git/alternatives-to-git-submodule-git-subtree). I'd prefer using git subtree instead.

Usage of the filter might become even easier if the YAML file was converted to a Lua table and integrated into the main script. Installation would be as simple as dropping the script in `$/HOME/.pandoc/filters`. But thanks to the good README, this is only a minor concern.

jgm:
Both submodule and subtree should be avoided, I think.

odkr:
Thanks! Also for the info about submodules. I wasn't aware of their disadvantages. I’ve made the changes you suggested and simplified the script.

tarleb:
Thank you for this and for your patience! Merging now.

Issue #46, opened 2019-03-02
Make tests more resistent against changes in pandoc


tarleb:
Many tests use golden files to check correctness of the produced output. This is fragile and breaks whenever the pandoc template changes for the given format. Tests should rather check for the existence of certain keywords in the output.

Filters with fragile tests:

- [x] scrlttr2
- [x] track-changes
- [x] table-short-captions

blake-riley:
I've updated the testing for table-short-captions.
Thanks for the examples in the other two filters.

PR #60, opened 2019-05-10
table-short-captions: replace golden files with expected strings

blake-riley:
Addresses #46 for table-short-captions


tarleb:
Great, thank you!

PR #63, opened 2019-06-24
Add sections.lua

nichtich:


cagix:
What is the difference compared to using the `--section-divs` option?

nichtich:
The benefit of my filter is that I practised coding in Lua only because I forgot or did not know about `--section-divs`. Thanks for the reminder!

PR #66, opened 2019-08-12
add a filter to change markdown h4 to latex output

efx:


tarleb:
Thank you for the contribution. The filter is a nice example of how filters can be used to modify specific parts of the document. However, it differs from most other filters in this repo in some key aspects, which makes me hesitant to add it in the current form:

- Normally, headers can be linked to via by specifying their ID. This is no longer possible for modified level 4 headers. That could lead to irritations for users who expect the scripts here to just work.
- Markup in the header (like emphasis or math) is lost during conversion.

Also, while many filters here are customizable via metadata fields, none require users to alter the filter's code. We aim at making filter usable without coding.

This might be a problem where a filter is not the best tool for the task, and which is easier to solve in LaTeX by configuring the `\paragraph` command. Can you try adding this to your document and see if it does what you need?

``` yaml
---
title: 4 levels of Markdown
header-includes: |
  ``` {=latex}
  \usepackage{titlesec}
  \titlespacing*{\paragraph}{0pt}{1ex}{-\parskip}
  \titleformat{\paragraph}[hang]
      {\normalfont\bfseries}
      {}
      {0pt}
      {}
  ```
---
```

efx:
Thanks for reviewing @tarleb !
I understand if this isn't a good candidate for inclusion in the repository so can just close the PR.
I do appreciate that answer but I cannot use it because of a warning. Pandoc's default template includes KOMA script's scrartcl which emits this warning when you attempt to use `titlesec`:
```
Class scrartcl Warning: Usage of package `titlesec' together
(scrartcl)              with a KOMA-Script class is not recommended.
(scrartcl)              I'd suggest to use the package only
(scrartcl)              if you really need it, because it breaks several
(scrartcl)              KOMA-Script features, i.e., option `headings' and
(scrartcl)              the extended optional argument of the section
(scrartcl)              commands.
(scrartcl)              Nevertheless, using requested
(scrartcl)              package `titlesec' on input line 107.
```

I did end up writing another filter to [center all images](https://tex.stackexchange.com/questions/46903/centering-with-includegraphics-not-with-beginfigure) when converting to LaTeX. Any interest in that one living here?

efx:
If you are on the Tex stack exchange site and can answer, I think your suggestion is the right answer for this question:
https://stackoverflow.com/questions/21198025/pandoc-generation-of-pdf-from-markdown-not-respecting-headers-formatting

tarleb:
For what it's worth, this is the best filter I could come up with, which forces the start of paragraphs following 4th level headers onto a new line. It's neither pretty nor very efficient, but seems to work ok.

```lua
local insert_paragraph_brake = false

function Header(h)
  if h.level == 4 then
    insert_paragraph_brake = true
  end
end

function Block(b)
  if not insert_paragraph_brake then return end
  insert_paragraph_brake = false
  if b.t == 'Para' then
    table.insert(b.content, 1, pandoc.RawInline('latex', '\\hfill{}\\\\{}'))
    return b
  end
end
```

I imagine that the image centering filter, if it is simple and well documented, could make a good addition to the examples in https://pandoc.org/lua-filters.html.

tarleb:
> If you are on the Tex stack exchange site and can answer, I think your suggestion is the right answer for this question:
https://stackoverflow.com/questions/21198025/pandoc-generation-of-pdf-from-markdown-not-respecting-headers-formatting

I added an answer.

bpj:
I added another answer, describing an alternative strategy.  More hacky but with zero extra dependencies:

https://stackoverflow.com/a/57525499/1640286

PR #68, opened 2019-08-17
add odt support in pagebreak lua filter

### LINKED ISSUE #67: pagebreak is not working with ODT document
I think there is some missing piece to make pagebreak filter work with ODT document, even following the readme. 

`pagebreak.odt` is not used and format `odt` is not known by lua filter.

cderv:
This fixes #67 by adding odt format in the known format for this filter.


tarleb:
Awesome, thanks!

PR #70, opened 2019-08-28
center images filter

efx:
Center images when converting from markdown to LaTeX
Following up from my comment in https://github.com/pandoc/lua-filters/pull/66

### DIFF
--- a/centerimage/centerimage.lua
+++ b/centerimage/centerimage.lua
@@ -1,8 +1,7 @@
 -- wrap images with a centering LaTeX snippet
 -- https://tex.stackexchange.com/questions/46903/centering-with-includegraphics-not-with-beginfigure
 function Image(elem)
-	local List = require 'pandoc.List'
-  return List:new{
+  return {
 		pandoc.RawInline('latex', '{\\centering'),
 		elem,
 		pandoc.RawInline('latex', '\\par}')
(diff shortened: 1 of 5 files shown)


tarleb:
Thank you for this. The small size and simplicity of this makes it suitable to be an example in the official Lua filter docs. See https://github.com/jgm/pandoc, file `doc/lua-filters.md`.

@jgm, what do you think about this? Format-specific output modifications seem to be a main use-case for filters, but we don't have a good example for that yet. Also, the example "Default metadata file" seems much less relevant now that we have `--metadata-file`, and we could replace it.

centerimage/centerimage.lua
```diff
@@ -0,0 +1,10 @@
+-- wrap images with a centering LaTeX snippet
+-- https://tex.stackexchange.com/questions/46903/centering-with-includegraphics-not-with-beginfigure
+function Image(elem)
+	local List = require 'pandoc.List'
+  return List:new{
```
tarleb: No need to use `List` here. The List type exists to make it easier to operate on Lua tables which represent lists, but it is not required to return a List type. Since we are not using any List features here, this could be shortened to

``` lua
return {
  pandoc.RawInline('latex', '{\\centering'),
  elem,
  pandoc.RawInline('latex', '\\par}')
}
```

centerimage/centerimage.lua
```diff
@@ -0,0 +1,10 @@
...
+		pandoc.RawInline('latex', '{\\centering'),
+		elem,
+		pandoc.RawInline('latex', '\\par}')
+  }
```
tarleb: This file uses both tabs and spaces for indentation, which can be very confusing. Most other filters in this repo use two spaces.

tarleb: yes, good idea!

tarleb:
Closing in favor of jgm/pandoc#6004

PR #73, opened 2019-09-06
Use absolute directories for temp files given to inkscape.

briansniffen:
Inkscape on Mac has a weird relocation problem:
/usr/local/bin/inkscape will typically be a script to cd into a
different directory.  Relative paths break.  I know this works on
POSIX-like systems, but I'm not so sure about Windows.

### DIFF
--- a/diagram-generator/diagram-generator.lua
+++ b/diagram-generator/diagram-generator.lua
@@ -93,9 +93,10 @@ end
 local function tikz2image(src, filetype, additionalPackages)
 
     -- Define file names:
-    local outfile = string.format("./tmp-latex/file.%s", filetype)
-    local tmp = "./tmp-latex/file"
-    local tmpDir = "./tmp-latex/"
+    local cwd = os.getenv("PWD")
+    local outfile = string.format("%s/tmp-latex/file.%s", cwd, filetype)
+    local tmp = string.format("%s/tmp-latex/file", cwd)
+    local tmpDir = string.format("%s/tmp-latex/", cwd)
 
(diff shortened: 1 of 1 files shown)


jgm:
Can you please give some more detailed information about what the problem is, preferably with a minimal example we can use to reproduce it?

briansniffen:
Inkscape on Mac installs into an “app bundle” instead of scattered across /usr. To compensate, it uses this program as /usr/local/bin/inkscape: https://gitlab.com/inkscape/inkscape/blob/INKSCAPE_0_92_3/packaging/macosx/Resources/bin/inkscape

It works well enough to start the GUI, but doesn’t deal well with relative paths: “.” is off in /Applications/Inkscape.app/Contents/ by the time it’s parsed. 

Using absolute paths works for me. I’m told an environment variable of “%CD%” is often set on Windows, but I don’t really have a dev machine for that. 

To reproduce this, I guess you get a Mac, install mactex, Xcode, and homebrew, and `brew cask install Inkscape`, and then `make` to run the included tests. Something will generate error messages about not finding the temporary PDF files emitted from latex—that’s Inkscape looking in the entirely wrong place. 
-- 
Brian Sniffen

tarleb:
This somehow fell entirely off my radar. I'm very sorry for the delay.

We've added a `pandoc.system` module in pandoc 2.7.3: the module provides functions to get the current working directory and to create temporary directories. This could solve the portability problems.

tarleb:
I've fixed this while also completely changing the way temporary files are handled. Thanks for bringing it up.

PR #74, opened 2019-10-15
Add list indices to institutes

jkrenzer:
Problem: We get the institutes per author as numeric indices, which is fine. But as we cannot enumerate in pandoc template $for loops, it is not trivial to link or reference against the institutes. Basically we cannot get a hold of the institutes index in pandoc template language.

Solution: Adding the index of the institute to the institute-object helps in getting this references straight. Now we can write, for example, a html-template and make use of anchor-links and id's.

### DIFF
--- a/scholarly-metadata/expected.md
+++ b/scholarly-metadata/expected.md
@@ -22,10 +22,13 @@ author:
 institute:
 - address: '23 Science Street, Eureka, Mississippi, USA'
   id: fosg
+  index: 1
   name: Formatting Open Science Group
 - id: fop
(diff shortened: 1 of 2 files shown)


jgm:
Note that in the next version of pandoc you will be able to access the indices in pandoc's templates.

jkrenzer:
That is good news, as I prefer to use templates for document-composition whenever possible. Thank you.

But still I would propose explicitly exposing attributes which we use for references.

tarleb:
Thanks, this seems like a good idea.

PR #79, opened 2019-12-11
Add lilypond filter

cole-miller:
This filter processes LilyPond source code into images of musical notation. See the README for a detailed description and the test files for some examples of how it's used.


cole-miller:
It looks like CI is failing due to a problem with another filter.

lilypond/lilypond.lua
```diff
@@ -0,0 +1,195 @@
...
+local SPECIAL_ATTRIBUTES = {
+        ["ly-caption"] = true,
+        ["ly-name"] = true,
+        ["ly-resolution"] = true,
+        ["ly-title"] = true
+      }
+
+-- pandoc.system.with_temporary_directory had a different (undocumented)
+-- name in the 2.7.3 release.
+-- Incidentally, the "right" way to do this version check is with a `Version'
```
tarleb: Huh, that's weird; `pandoc.types` should be available.

lilypond/lilypond.lua
```diff
@@ -0,0 +1,195 @@
...
+-- object, but despite what the documentation says `pandoc.types' doesn't
+-- seem to be available in v2.7.3.
+local with_temporary_directory = tostring(PANDOC_VERSION) == "2.7.3"
```
tarleb: We use some Lua magic so make `PANDOC_VERSION` simple to use, so `PANDOC_VERSION < '2.8'` would do what one would expect. The "2.8" is transformed into a Version object before the comparison is done.

Your code is perfectly fine, I mention it mostly because I think it's a cool hack :man_shrugging:

lilypond/lilypond.lua
```diff
@@ -0,0 +1,195 @@
...
+          )
+          pandoc.pipe("cp", {fullname, whither}, "")
+        end
+      )
+    end
+  )
+  return whither .. "/" .. fullname
+end
+
+function make_relative_path(to, from)
```
tarleb: It's probably worth to include such a function in pandoc.system. The script would become more portable. Would you raise an issue in the pandoc repo?

lilypond/lilypond.lua
```diff
@@ -0,0 +1,195 @@
...
+  return pandoc.pipe(
+           "realpath",
+           {"--relative-to=" .. from, to},
+           ""
+         ):gsub("\n", "")
+end
+
+local function process_lilypond(elem)
+  if elem.classes:includes("lilypond") then
```
tarleb: One could reduce nesting a little by combining these two if statements:

``` lua
if (not elem.classes:includes("lilypond")
    or elem.classes:includes("ly-norender")) then
  return elem
end

-- rest of the function goes here
```

lilypond/lilypond.lua
```diff
@@ -0,0 +1,195 @@
...
+    if elem.classes:includes("ly-norender") then
+      return elem
+    end
+
+    -- Are we dealing with an inline code element or a code block?
+    local inline = elem.tag == "Code"
```
tarleb: That's clever

lilypond/lilypond.lua
```diff
@@ -0,0 +1,195 @@
...
+  end
+end
+
+-- Update `OPTIONS' based on the document metadata. There's a bit of
+-- munging because string options are parsed as (pandoc-flavored) Markdown
+-- and we have to dig out the raw Lua string.
+local function meta_transformer(md)
+  local ly_block = md.lilypond or {}
+
+  OPTIONS.image_directory = ly_block.image_directory
```
tarleb: `pandoc.utils.stringify` could come in handy here.

cole-miller: Thanks, that slipped my mind.

cole-miller:
Thanks for your feedback. I've opened a [feature request](https://github.com/jgm/pandoc/issues/6001) in the pandoc repo as you suggested (just barely missed getting #6000!).

tarleb:
Thanks again, this is a great addition!

tolot27:
How about LaTeX? Can it be simply converted to the lilypond environment?

cole-miller:
@tolot27 I don't know what LaTeX syntax elements pandoc translates into `Code` and `CodeBlock`, if any, or how you'd specify element attributes. You might be better off writing in Markdown and translating to PDF with pandoc, which passes through LaTeX as an intermediate step.

Issue #69, opened 2019-08-18
section-refs duplicates references


hadrienj:
Hi,

Thank you very much for the work on the filter section-refs! I tried to use it but encounter a problem: the references are duplicated. If I cite a source 9 times in the chapter, it will appear 9 times in the bibliography.

<img width="739" alt="Screenshot 2019-08-18 at 13 58 56" src="https://user-images.githubusercontent.com/6381368/63224171-b10fe000-c1c0-11e9-8b84-7d69f65f63a1.png">

Any idea how to fix this?

Thanks!

bpj:
Well table with reference keys as table keys which are set to a true value
if the reference has already been seen and checking/setting it
appropriately ought to do the trick.

That code has some other issues too. All helper functions are declared as
global which might not be a good idea, but what's worse it seems it
sometimes forgets to check that metadata values are the right type before
just using them as the content of other elements. Hardly a good idea in
code meant to be used by others!

Den sön 18 aug. 2019 14:04hadrienj <notifications@github.com> skrev:

chemcoder-2020:
I'm running into the same problem. Did you ever figure out this issue?

chemcoder-2020:
Ok so I modified the code for `section-refs.lua` as follows:

First I added before the `get_partial_refs` function the following helper function:
```
function table.contains(table, element)
   for _, value in pairs(table) do
     if value == element then
       return true
     end
   end
   return false
 end
```
Then I added an if statement at the end of the `get_partial_refs` function:
```
if not table.contains(refs, nr[2]) then
      table.insert(refs, nr[2])
end
```

Overall the filter file becomes:
```
function is_ref_div (blk)
   return (blk.t == "Div" and blk.identifier == "refs")
end

function is_ref_header (blk)
   return (blk.t == "Header" and blk.identifier == "bibliography")
end

function get_all_refs (blks)
   for _, b in pairs(blks) do
      if is_ref_div(b) then
	 return b.content
      end
   end
end

function remove_all_refs (blks)
   local out = {}
   for _, b in pairs(blks) do
      if not (is_ref_div(b) or is_ref_header(b)) then
	 table.insert(out, b)
      end
   end
   return out
end

-- We return a {number, ref} pair so we can sort in the individual
-- bibliographies.
function citation_to_numbered_ref (citation, all_refs)
   local div_id = "ref-" .. citation.id
   for i, d in ipairs(all_refs) do
      if d.t == "Div" and d.identifier == div_id then
[...]

PR #87, opened 2020-01-23
Fixing duplicate references for Issue#69

chemcoder-2020:
New pull request for fixing section-refs lua filter to remove duplicate references.
@tarleb #69

### DIFF
--- a/section-refs/section-refs.lua
+++ b/section-refs/section-refs.lua
@@ -6,10 +6,14 @@ function is_ref_header (blk)
    return (blk.t == "Header" and blk.identifier == "bibliography")
 end
 
+function is_first_header (blk)
+   return (blk.level == 1) -- identify if first header for resetting citations later
+end
+
 function get_all_refs (blks)
    for _, b in pairs(blks) do
       if is_ref_div(b) then
-	 return b.content
+         return b.content
       end
    end
 end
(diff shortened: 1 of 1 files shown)


section-refs/section-refs.lua
```diff
@@ -66,7 +74,9 @@ function get_partial_refs (blocks, all_refs)
 
    local refs = {}
```
tarleb: `table.contains` is equivalent to `pandoc.List.includes`. For idiomatic use, I'd suggest

``` lua
local refs = pandoc.List:new()
for _, nr in pairs(numbered_refs) do
  if not refs:includes(nr[2]) then
    table.insert(refs, nr[2])
  end
end

File: section-refs/section-refs.lua [hunk 2]
```diff
@@ -135,4 +145,4 @@ function Pandoc(doc)
       local output = add_section_refs(unreffed, lvl, refs_title, all_refs)
       return pandoc.Pandoc(output, doc.meta)
    end
-end
+end
```
tarleb: Unix convention is to have a newline as the last character of a text file.

tarleb:
The tests need adaption, as the sample file also contains a duplicate. I'm merging the code as is and will fix the tests. Thanks for your patience and the PR.

PR #100, opened 2020-07-13
Set a default for python_path.

### LINKED ISSUE #99: Using yaml-metadata to set the dotPath doesn't seem to work?
I am using pandoc 2.10 together with the current version of lua-filters/diagram-generator. I try to create diagrams using graphviz. But for one page I need the `neato`-binary from graphviz instead of `dot`. According to the documentation I should be able to set metadata `dotPath` to affect this. This works cool from the command line via `--metadata dotPath=/usr/bin/neato`. But if I try to add the information inside the file as follows

~~~~~~
---
dotPath: /usr/bin/neato
---
# Test a graphviz image.

~~~graphviz
graph test {
    A -- B;
    C -- B;
    D -- B;
    B -- E;
    C -- E;
}
~~~
~~~~~~

I just get an error message: `PandocLuaError "Error during function call: expected string, got 'table: 0x5e11780' (table)"`. It seems that pandoc returns a table with the string inside a `text` key. I am not sure if I am doing something wrong, or if maybe in pandoc the structure of metadata was changed somewhere to 2.10.  I tried to work around the problem, but I am either incompatible to the structure coming from the yaml-meta or coming from the commandline metadata parameter. Can anybody help?

jerri:
Fixes #99.

### DIFF
--- a/diagram-generator/diagram-generator.lua
+++ b/diagram-generator/diagram-generator.lua
@@ -32,7 +32,7 @@ local inkscape_path = os.getenv("INKSCAPE") or "inkscape"
 
 -- The Python path. In order to define a Python version per pandoc document,
 -- use the meta data to define the key "python_path".
-local python_path = os.getenv("PYTHON")
+local python_path = os.getenv("PYTHON") or "python"
 
 -- The Python environment's activate script. Can be set on a per document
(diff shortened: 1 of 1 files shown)


tarleb:
Thanks!

PR #103, opened 2020-07-17
Add filter revealjs-codeblock

sokotim:
Adds a filter that passes code block attributes to `revealjs` HTML output to enable reveal.js code presentation features like line numbers & highlights, step-by-step highlights and auto-animation.


tarleb:
Thank you for the PR! It will take me a little longer to review, but should be able to do so next week.

revealjs-codeblock/revealjs-codeblock.lua
```diff
@@ -0,0 +1,72 @@
...
+
+local function is_data_line_number_in_attributes(attributes)
+  for _, attribute in ipairs(attributes) do
+    if attribute[1] == 'data-line-numbers' then return true end
+  end
+  return false
+end
+
+function CodeBlock(block)
+  if FORMAT == 'revealjs' then
```
tarleb, line 34: This could go to the top of the file, like so:

``` lua
if FORMAT ~= 'revealjs' then
  return {}
end
```

revealjs-codeblock/revealjs-codeblock.lua
```diff
@@ -0,0 +1,72 @@
...
+local function is_numberlines_class(class)
+  return class == 'numberLines' or class == 'number-lines'
+end
+
+local function is_pre_tag_attribute(attribute)
+  return attribute == 'data-id'
+end
+
+local function is_data_line_number_in_attributes(attributes)
+  for _, attribute in ipairs(attributes) do
```
tarleb, line 27: Attributes come with a metatable which allows using the structure just like an associative table. E.g., this could be written as

``` lua
for key, value in pairs(attributes) do
    if key == 'data-line-numbers' then return true end
end
```

revealjs-codeblock/revealjs-codeblock.lua
```diff
@@ -0,0 +1,72 @@
...
+        if not is_data_line_number_in_attributes(block.attributes) then
+          table.insert(block.attributes, {'data-line-numbers', ''})
+        end
+      else
+        table.insert(css_classes, class)
+      end
+    end
+    if block.identifier ~= '' then
+      table.insert(pre_tag_attributes,
+                    string.format('id="%s"', block.identifier))
```
tarleb, line 50: This could be extracted in a separate function like `html_attr` or the like.

PR #105, opened 2020-07-27
Adds automatic shifting

St-Ex-savadenn:
Enables auto-shifting titles
With this feature all your files can begin with a level 1 heading. When included, it will shift accordingly to the previous title found.

Also adds a warning message if file is not found rather than crash


tarleb:
Sorry for letting this lay around for so long, and thank you for your patience.

This looks nice, and I'd like to merge it. Maybe we could add a little warning that the result may not be as expected in certain cases. E.g., some people might expect the include to be inserted at level 1, but it would be inserted at level 2 here:

```` markdown
# Title

:::{.wrapping-div}
## level 2
:::

```{.include}
file.md
```
````

Could you break the code changes up into two parts?  The warning message is a good idea and should be in its own commit, or even PR. The rational here is that, if we ever have to revert a feature for whatever reason, it is much easier to do so if only a single relevant commit has to be reverted.

Finally a request about style: the new code mixes camelCase with snake_case; we try to keep things consistent by following the [Olivine labs style guide](https://github.com/Olivine-Labs/lua-style-guide). The suggested naming style for variables is snake_case.

Cheers!

**Comment by St-Ex-savadenn** [comment]
Thanks for your advice.
Here my to do list :
* [ ] Add a PR with the warning for not found file
* [ ] Review styling on this PR
* [ ] Add a better explanation in the readme on what is the default level of injection and how to specify the desired one

**Comment by St-Ex-savadenn** [comment]
@tarleb MR is ready. Review it when you can, thx

File: include-files/include-files.lua [hunk 1]
```diff
@@ -6,6 +6,20 @@
 -- pandoc's List type
 local List = require 'pandoc.List'
 
+--- Get include auto mode
+local include_auto = false
+function get_vars (meta)
+  if meta['include-auto'] then
```
tarleb, line 12: In case you like [code golf](https://en.wikipedia.org/wiki/Code_golf):

``` lua
function get_vars (meta)
  include_auto = not not meta['include-auto']
end
```

include-files/include-files.lua
```diff
@@ -32,9 +46,21 @@ function transclude (cb)
 
   -- Markdown is used if this is nil.
   local format = cb.attributes['format']
-  local shift_heading_level_by =
-    tonumber(cb.attributes['shift-heading-level-by'])
 
+  -- Attributes shift headings
+  local shift_heading_level_by = 0
+  local shift_input = cb.attributes['shift-heading-level-by']
+  if shift_input then
```
tarleb, line 53: Just for fun, some code golf again.
``` lua
local shift_heading_level_by = shift_input and tonumber(shift_input) or
  (include_auto and last_heading_level or 0)
```
Probably wouldn't help readability though.

tarleb (approved):
Thanks! Merging...

include-files/include-files.lua
St-Ex-savadenn: I do like this one, use it all the time in JS 

```js
let foo = !!bar
```

tarleb: Feel free to submit more PR's if you want to fine-tune

PR #110, opened 2020-09-22
add escape char for '<' and ’>'

code-in-cpp:


code-in-cpp:
This update is to fix below description

Content added inside of a <code> block is parsed as HTML by the web browser. If you have HTML characters (<>) in your code you will need to escape them ($lt; $gt;).

code-in-cpp:
> _No description provided._

tarleb:
Thanks for the PR! Why did you close it? Seems like a good fix for a real issue.

code-in-cpp:
I am not familiar with PR and raised 2 same PR at one time, so I closed one

PR #117, opened 2020-10-08
[include-files] Handles not found

St-Ex-savadenn:
Warn user if file to include is not found and skip it rather than crash


tarleb:
Thanks!

PR #118, opened 2020-10-11
math2svg.lua filter

stroobandt:
This Lua filter converts LaTeX math to MathJax generated SVG in any of the available MathJax fonts.


tarleb:
Thank you for the PR. The documentation looks nice, tests are good, and
the filter seems useful.

There are a couple of questions and issues which we should check to make
the filter maintainable as well as useful for other users.

A problem I see is that the docs are focused on Ubuntu/Debian and the
path to `tex2svg` is hard-coded in the filter. We should either have
clear instructions for all major platforms, or just link to instructions
provided elsewhere.

The lua-filters repo is under the MIT license; I'd like to prevent a
jumble of filter licenses. For that reason we can't include code
licensed under the GPL-3.

The addition of the additional LaTeX commands is opinionated; can we
find a way to make that more general?

Could you also add an explanation to the docs why only display math
equations are handled, but not inline math? I assume it is because of
baseline alignment issues? How does Mathjax handle that normally?

The filter currently adds raw SVG code to all formats, which is most
likely not what we want. Maybe we could store the images in the mediabag
in these cases?

Cheers!

stroobandt:
I need some help here with the test infrastructure.
For the Travis CI build to be successful, the following programs are required:

stroobandt:
Thank you for the very constructive comments.
I will work on these in the coming days.

stroobandt:
Dear Albert,

I think I have been able to attend to all your concerns:

+ The path to `tex2svg` has been isolated in a configuration variable. A comment explains how to locate this executable.
+ The filter has been licensed under the MIT license.
+ The `newcommands` variable is now empty by default. An example is provided in the comments.
+ When the output format is not HTML, the SVG is now inserted into the [mediabag](https://pandoc.org/lua-filters.html#module-pandoc.mediabag).
+ The program can now also convert inline math to SVG. The type of math that gets converted, display versus inline, is determined by two new configuration variables. Inline math conversion to SVG is switched off by default and falls back to MathML. The reasons for this are explained in the comments.
+ The name of the filter changed to `math2svg.lua` to reflect the changes in the point above.

math2svg/.gitignore
```diff
@@ -0,0 +1 @@
+*.swp
```
tarleb, line 1: This is best put into the global `$HOME/.gitignore` file.

math2svg/README.md
```diff
@@ -0,0 +1,77 @@
+# Description
+
+This [Lua filter](https://pandoc.org/lua-filters.html)
+for [Pandoc](https://pandoc.org/)
+converts [LaTeX math](https://en.wikibooks.org/wiki/LaTeX/Mathematics)
+to [MathJax](https://www.mathjax.org/) generated
+[scalable vector graphics (SVG)](https://en.wikipedia.org/wiki/Scalable_Vector_Graphics)
+in any of the [available MathJax fonts](https://docs.mathjax.org/en/latest/output/fonts.html).
+
+This is useful when a CSS paged media engine cannot process complex JavaScript
```
tarleb, line 10: It seems worthwhile to highlight the privacy aspect of this filter. If just no JS is the goal, then a simple `pandoc --webtex --self-contained` would be enough.

math2svg/README.md
```diff
@@ -0,0 +1,77 @@
+# Description
```
tarleb, line 1: Convention in other filters is to put the filter name as *the* level 1 header, and all other headings at level 2 or below.

math2svg/README.md
```diff
@@ -0,0 +1,77 @@
...
+No Internet connection is established when creating MathJax SVG code using
+the `tex2svg` command of [`mathjax-node-cli`](https://github.com/mathjax/mathjax-node-cli).
+Hence, formulas in SVG can be created offline and will remain private.
+For code auditing, see also:
+
+- <https://github.com/mathjax/MathJax-node>
+- <https://github.com/pkra/mathjax-node-sre>
+
+
+# Copyright
```
tarleb, line 47: A common practice is to name this section "License", without further subsections.

math2svg/math2svg.lua
```diff
@@ -0,0 +1,137 @@
+--  DESCRIPTION
```
tarleb, line 1: This is a lot of information duplicated from the README. Most filters just contain a short (single paragraph) description, as well as a copyright/licensing notice.

math2svg/math2svg.lua
```diff
@@ -0,0 +1,137 @@
...
+
+--  CONTACT
+--
+--    $ echo c2VyZ2VAc3Ryb29iYW5kdC5jb20K |base64 -d
+
+
+--  Enter here the full path to the tex2svg binary of mathjax-node-cli.
+--  The full path can be found with the following command:
+--    $ which tex2svg
+local tex2svg = '/usr/local/lib/node_modules/mathjax-node-cli/bin/tex2svg'
```
tarleb, line 76: One of the guiding principles for the filters here is that it should not be necessary to modify any filter itself to adapt it to (most) alternative use-cases and environments. E.g., the `diagram-generator` filter uses environment variables to allow overwriting of default variables. See <https://github.com/pandoc/lua-filters/blob/master/diagram-generator/diagram-generator.lua#L33-L35>. It is even possible to override the program path using metavalues, implemented [here](https://github.com/pandoc/lua-filters/blob/master/diagram-generator/diagram-generator.lua#L88-L90). That last part is nice, but more of a bonus.

math2svg/math2svg.lua
```diff
@@ -0,0 +1,137 @@
...
+      svg = '<div class="math display">' .. svg .. '</div>'
+      return pandoc.RawInline(FORMAT, svg)
+    else
+      local filename = pandoc.sha1(svg) .. '.svg'
+      return pandoc.mediabag.insert(filename, 'image/svg+xml', svg)
+    end
+
+  elseif elem.mathtype == 'InlineMath' and inline2svg then
+
+    local svg = pandoc.pipe(tex2svg, {'--speech=false', '--font', font, newcommands .. elem.text}, '')
```
tarleb, line 125: This is the same line as above. Could this be moved outside of the `if` clause to reduce duplication?

math2svg/math2svg.lua
```diff
@@ -0,0 +1,137 @@
...
+
+    if FORMAT:match '^html.?' then
+      svg = '<span class="math inline">' .. svg .. '</span>'
+      return pandoc.RawInline(FORMAT, svg)
+    else
+      local filename = pandoc.sha1(svg) .. '.svg'
+      return pandoc.mediabag.insert(filename, 'image/svg+xml', svg)
```
tarleb, line 132: This will insert the image into the mediabag, but the function returns `nil`. As a result pandoc will leave the math element in place unchanged. We should return a `pandoc.Image` element with the new `filename` for the image to show up in the result.

tarleb (commented):
Thank you for the quick update. There are a few things which I forgot to include in my first comment, and was not specific enough about others. See below for a couple of inline comments.

Cheers!

math2svg/.gitignore
stroobandt: `.gitignore` was removed.

math2svg/README.md
stroobandt: Good point. I have added the following line to the description:

> No Internet connection is required for SVG conversions, resulting in absolute privacy.

However, for me personally, the main pain points with `--webtex` are:

1. no font selection
2. no distinction between display and inline math

math2svg/README.md
stroobandt: Done. I changed the heading levels and added the filter name.

math2svg/README.md
stroobandt: Done. I changed the heading to License.

math2svg/math2svg.lua
stroobandt: Done. A `pandoc.Image` is now returned for output formats other than HTML.

math2svg/math2svg.lua
stroobandt: Done. I have refactored the code.

math2svg/math2svg.lua
```diff
@@ -0,0 +1,137 @@
...
+local inline2svg  = false
+--  The fallback is MathML if pandoc is executed with the --mathml argument.
+--  MathML output gets generated much faster than SVG output.
+--  Moreover, MathML is better suited to InlineMath as line heights are kept small.
+
+
+--  Any additional LaTeX math commands can be defined here.
+--  For example:
+--  local newcommands = '\\newcommand{\\j}{{\\text{j}}}\\newcommand{\\e}[1]{\\,{\\text{e}}^{#1}}'
+local newcommands = ''
```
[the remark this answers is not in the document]
stroobandt: Done. The `newcommands` variable was entirely removed from the code.

math2svg/math2svg.lua
stroobandt: I defer. All to often, the filter `.lua` is the only file that ends up on the server.
Then when maintenance comes around, a couple of months later, this comment section inside the `.lua` file happens to be the only reminder of what this file is for.

Comments never hurt; the lack of any does. I speak from experience.

math2svg/README.md
```diff
@@ -0,0 +1,77 @@
...
+See: <https://www.print-css.rocks> for information about CSS paged media,
+a [W3C standard](https://www.w3.org/TR/css-page-3/).
+
+The filter also allows to define additional LaTeX commands.
+
+
+# Requires
+
+```bash
+$ sudo apt install pandoc nodejs npm
```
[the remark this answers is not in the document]
stroobandt: Install instructions are now provided for all major GNU/Linux distributions, as well as macOS and Windows.

math2svg/math2svg.lua
tarleb: Wouldn't a reference to the main repo suffice for that purpose?

The reason that I'm against this is that it breaks the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle. Duplicates tend to diverge over time, people will forget to update, and all well-done updates cost twice the time. An outdated comment is often worse than no comment at all, as it spreads disinformation and confusion. I've been guilty of that myself and am not particularly keen to repeat that experience.

stroobandt: OK. I reduced the description to a single paragraph and added a hyperlink.

math2svg/math2svg.lua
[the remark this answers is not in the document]
stroobandt: That is a neat little trick!
Reading from the [pandoc documentation](https://pandoc.org/MANUAL.html), one would not think this would even be possible at all. After all, my output format is HTML and not LaTeX; well, apart from the formulas. A text search reveals that the pandoc manual is really scarce and scattered about `header-includes`.

In my eternal quest to separate format from content, I would like to achieve exactly the same from the `makefile`.
I was hoping `-H FILE` would do that, as [suggested in the manual](https://pandoc.org/MANUAL.html#variables-set-automatically). However, I somehow cannot get it to work. Can you help me with that?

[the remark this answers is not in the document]
stroobandt: Neither `--metadata=header-includes: …` nor `--metadata-file=FILE.yaml` work for me. I got stuck, please help.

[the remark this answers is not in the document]
stroobandt: OK, I found how to do it in [_Boilerplating Pandoc for Academic Writing_](https://www.soimort.org/notes/161117/).

As mentioned before, I experience the pandoc manual as severely lacking and especially confusing on the subject of `header-includes`. Quite a bit of time was lost at something that it is really simple.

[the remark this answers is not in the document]
stroobandt: For your information, I wrote up my findings in this [TeX StackExchange answer](https://tex.stackexchange.com/a/566707/26348).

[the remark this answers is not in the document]
tarleb: Please raise this as an issue on pandoc's issue tracker. Ideally, it should contain a short summary of what you tried to do, and why the available information was insufficient for the task at hand. A link to this conversation might be enough.

The reason that none of the other options worked is that the definition must be parsed by the Markdown reader. This is not the case unless the definitions are placed in the main document. Howerver, the `\newcommand` doesn't have to be put into the `header-includes` frontmatter (although that is usually the most natural place). The raw LaTeX can occur directly in the text. Since pandoc concatenates multiple input files, and allows to include `\newcommand` in the text, one could write

    pandoc --from=markdown my-newcommand-definitions.tex my-actual-document.md

where `my-necommand-definitions.tex` would contain all definitions

    \newcommand{\j}{{\text{j}}}
    ...

and get the desired result.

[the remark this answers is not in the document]
stroobandt: Done. The whole `newcommands` code was removed.

stroobandt:
TODO:

- [ ] `README.md`: Add `--metadata` key and value information.

math2svg/math2svg.lua
stroobandt: Done. The filter is now fully configurable over `--metadata` key value pairs.
However, information about this still needs to be added to the `README.md` documentation.

math2svg/math2svg.lua
```diff
@@ -0,0 +1,137 @@
...
+--    $ which tex2svg
+local tex2svg = '/usr/local/lib/node_modules/mathjax-node-cli/bin/tex2svg'
+
+
+--  Supported MathJax fonts are: https://docs.mathjax.org/en/latest/output/fonts.html
+local font = 'Gyre-Pagella'
+
+
+--  Indicate with true or false whether DisplayMath and InlineMath should be converted to SVG.
+local display2svg = true
```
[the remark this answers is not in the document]
stroobandt: Done. The filter is now fully configurable over `--metadata` key value pairs.
However, information about this still needs to be added to the `README.md` documentation.

stroobandt (commented):
Dear Albert,

All your pending issues have been addressed and the documentation has been completed.
Could you, please, consider finalising this pull request? Thank you.

tarleb:
Thank you. This is looking really good now!

The tests currently fail, partly because I need to adapt existing tests to the latest pandoc release, and also because `tex2svg` isn't being installed on the CI system. You could try to do the latter yourself by editing `.travis.yml` if you'd like; otherwise just let me know and I'll have a look once the other tests are fixed. I'll be holding off merging until we have all lights green again.

Some tweaks we could look at until then:

- Couldn't we set the default for `math2svg_path` to just `tex2svg`? This would allow users to simply add the executable into their `PATH` and would make the filter truly platform independent. What do you think about naming that parameter `tex2svg_path` to make it more obvious what it's used for?
- Some lines are very long, both in the filter and the readme. Lines should stay below 80 chars if possible. If your editor supports it, I can very much recommend <https://editorconfig.org> which helps to keep things consistent.
- Yesterday I found <https://github.com/jgm/pandoc-tex2svg>. Maybe we should mention in the docs that your filter can be used as a more configurable alternative?
- If we were to set `math2svg_inline2svg` to `true` by default, then the filter could serve as a full drop-in replacement for [pandoc-tex2svg](https://github.com/jgm/pandoc-tex2svg). Personally, I'd also find this more consistent, but it is your choice.
- I remembered that HTML does not allow `<div>` elements within `<p>` elements. The filter currently produces invalid markup and should probably use `<span>` for both "display" and "inline" elements. We could think about adding `style="display:block"` to "display" spans, but maybe it's better to leave that to a global CSS rule.

stroobandt:
> You could try to do the latter yourself by editing `.travis.yml` if you'd like

Done. It seems to be working. I also cleaned up a bit the YAML indenting as per Travis CI documentation.

> * Couldn't we set the default for `math2svg_path` to just `tex2svg`? This would allow users to simply add the executable into their `PATH` and would make the filter truly platform independent. What do you think about naming that parameter `tex2svg_path` to make it more obvious what it's used for?

Done. The key is now called `math2svg_tex2svg`. When absent, the filter will now just try to execute `tex2svg`; rendering it truly platform independent.

> * Some lines are very long, both in the filter and the readme. Lines should stay below 80 chars if possible.

Reducing the lines below 80 columns is simply not possible because some of the hyperlinks in `README.md` are longer than that.
However, I was able to reduce `README.md` and `math2svg.lua` to below 120 columns, which also happens to be a terminal standard. (Hey, we are not [1984](https://en.wikipedia.org/wiki/Nineteen_Eighty-Four) anymore! — Although it certainly feels like it nowadays.)

> * Yesterday I found https://github.com/jgm/pandoc-tex2svg. Maybe we should mention in the docs that your filter can be used as a more configurable alternative?

To set the record straight: More than a year before John was even considering this, [others and myself were already playing with math SVG](https://groups.google.com/forum/#!topic/pandoc-discuss/TbOlj9qv94Y). Credits for suggesting the current solution go out to Nikolay "Lierdakil" Yakimov.

Anyhow, no, I do not think this needs mentioning. What I do think is that our Lua filters deserve **BETTER MARKETING.**
I will open a separate issue about this for discussion. Currently. Lua filters look more like [an afterthought at the bottom of the Extras page](https://pandoc.org/extras.html), when they are actually easier to maintain than Haskell filters. When one eventually lands on [this `README.md`](https://github.com/pandoc/lua-filters/), it becomes a _GitHub dumpster dive_ to find out which Lua filter does what. I would suggest adding one paragraph descriptions to the general [`README.md`](https://github.com/pandoc/lua-filters/) and do something similar on the [Extras page](https://pandoc.org/extras.html).

> * If we were to set `math2svg_inline2svg` to `true` by default, then the filter could serve as a full drop-in replacement for [pandoc-tex2svg](https://github.com/jgm/pandoc-tex2svg). Personally, I'd also find this more consistent, but it is your choice.

Won't fix. In the `README.md` I promise the user _sensible_ defaults. Even [John complains about the filter being slow](https://github.com/jgm/pandoc-tex2svg) when both display and inline math are being converted:

> _"The filter is rather slow, and it adds significantly to file size, but the resulting HTML renders quickly and does not depend on an internet connection or JavaScript."_

> * I remembered that HTML does not allow `<div>` elements within `<p>` elements. The filter currently produces invalid markup and should probably use `<span>` for both "display" and "inline" elements. We could think about adding `style="display:block"` to "display" spans, but maybe it's better to leave that to a global CSS rule.

Work in progress. I already fixed it with a `<span class="math display">`. However, since [I eat my own dog foot](https://hamwaves.com/lpda/en/lpda.a4.pdf), I discovered this is not ideal for CSS paged media processing. I will run some tests to see what works best, whilst keeping the HTML valid.

stroobandt:
I opened #121 concerning better Lua filter marketing.

math2svg/math2svg.lua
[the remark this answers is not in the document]
stroobandt: FYI, I opened Pandoc issue <https://github.com/jgm/pandoc/issues/6757>

tarleb:
The additional, non-essential changes to `.travis.yml` broke the config. PRs should only have one topic. Changing indentation in a config file is a separate concern and should be treated accordingly.

The 80 char-per-line rule is rooted in two principles: (1) The README should be easy to read, no matter whether viewed as rendered HTML in a browser, or as plaintext on a console. (2) research showed that text is easiest to read if there a 60 to 66 characters per line. 80 is still acceptable, anything above 100 is seriously bad. If you don't like how links look when forcing shorter lines, use [reference links](https://commonmark.org/help/tutorial/07-links.html).

math2svg/README.md
```diff
@@ -0,0 +1,229 @@
...
+a [W3C standard][w3c].
+
+No Internet connection is required when generating or viewing SVG formulas,
+resulting in both absolute privacy and offline, standalone robustness.
+
+
+## Requirements
+
+First, use the package manager of your operating system to install `pandoc`,
+`nodejs` and `npm`. `brew` and `choco` are recommended package mangers for
```
cderv, line 24: Just reading through this a saw this typo
```suggestion
`nodejs` and `npm`. `brew` and `choco` are recommended package managers for
```

stroobandt: Thanks!

tarleb:
We had to switch from Travis CI, so now all tests are run in GitHub Actions. This led to a merge conflict which has to be resolved before we can finally merge. The additional packages now need to be added in `.tools/Dockerfile`. Could you update the PR?

stroobandt:
@tarleb Dear Albert, I have modified `.tools/Dockerfile`, but apparently I cannot test it because of lacking write permissions.

> Only those with write access to this repository can merge pull requests.

The packages `nodejs` and `npm` are required for testing.
There is also an additional `RUN` statement to download an `npm` package.

stroobandt:
All open issues have been resolved.
If an administrator can take care of the suggested `.tools/Dockerfile`,
this new Lua filter should be ready to go.

tarleb:
Thanks!

PR #123, opened 2020-10-18
Add asciidoc format to pagebreak.lua

niszet:
For asciidoc and asciidoctor format, pagebrek.lua is updated.
Syntax is from AsciiDoc Syntax Quick Reference.
https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#breaks
Currently, RawBlock in AsciiDoc Writer does not add new lines. So, this update have "\n\n" to separate paragraphs.

### DIFF
--- a/pagebreak/pagebreak.lua
+++ b/pagebreak/pagebreak.lua
@@ -23,13 +23,13 @@ end
 
 --- configs – these are populated in the Meta filter.
 local pagebreak = {
+  asciidoc = '<<<\n\n',
   epub = '<p style="page-break-after: always;"> </p>',
   html = '<div style="page-break-after: always;"></div>',
   latex = '\\newpage{}',
   ms = '.bp',
   ooxml = '<w:p><w:r><w:br w:type="page"/></w:r></w:p>',
-  odt = '<text:p text:style-name="Pagebreak"/>',
-  asciidoc = '<<<\n\n'
+  odt = '<text:p text:style-name="Pagebreak"/>'
 }
 
 local function pagebreaks_from_config (meta)
--- a/pagebreak/pagebreak.lua
+++ b/pagebreak/pagebreak.lua
@@ -28,7 +28,8 @@ local pagebreak = {
   latex = '\\newpage{}',
   ms = '.bp',
   ooxml = '<w:p><w:r><w:br w:type="page"/></w:r></w:p>',
-  odt = '<text:p text:style-name="Pagebreak"/>'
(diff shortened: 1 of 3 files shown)


pagebreak/pagebreak.lua
```diff
@@ -58,6 +59,10 @@ local function newpage(format)
     return pandoc.RawBlock('html', pagebreak.html)
   elseif format:match 'epub' then
     return pandoc.RawBlock('html', pagebreak.epub)
+  elseif format:match 'asciidoc' then
+    return pandoc.RawBlock('asciidoc', pagebreak.asciidoc)
+  elseif format:match 'asciidoctor' then
```
tarleb, line 64: I think the second elseif can be omitted, as `asciidoc` will also match `asciidoctor`.

pagebreak/pagebreak.lua
```diff
@@ -27,7 +27,8 @@ local pagebreak = {
   html = '<div style="page-break-after: always;"></div>',
   latex = '\\newpage{}',
   ooxml = '<w:p><w:r><w:br w:type="page"/></w:r></w:p>',
-  odt = '<text:p text:style-name="Pagebreak"/>'
+  odt = '<text:p text:style-name="Pagebreak"/>',
+  asciidoc = '<<<\n\n'
```
tarleb, line 31: We don't do this yet, but maybe we should sort all formats alphabetically. Now that we get support for more and more formats, maintaining an alphabetical order could help us to keep things clean in the future.

tarleb (commented):
Thanks for this! I will merge once I fixed the tests.

pagebreak/pagebreak.lua
niszet: Oh, OK, I will omit this line, `asciidoctor`.

pagebreak/pagebreak.lua
niszet: OK, I understand and it is a good idea !  I will sort these items.
And, I notice `asciidoc` can have an additional test. I will add it. Thank you for reviewing!

niszet:
I tried to add a test for `asciidoc`.

tarleb:
Sorry, I merge another PR which caused a merge conflict. Could you rebase to the current master? We should probably also mention asciidoc support in the README.

niszet:
OK, no problem. I will rebase and update.

niszet:
I rebased. If any problem remains, I will update again.

tarleb:
Thanks!

PR #124, opened 2020-10-18
Add groff ms format support to pagebreak.lua.

a-vrma:
The way {g,t}roff works is that the macro, `.bp`, has to be on a line by itself. Since we are creating a RawBlock, I think that is implied.

### DIFF
--- a/pagebreak/Makefile
+++ b/pagebreak/Makefile
@@ -1,6 +1,11 @@
 DIFF ?= diff --strip-trailing-cr -u
 
-test:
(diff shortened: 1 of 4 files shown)


tarleb:
Thanks! I'll merge once I have fixed the tests.

tarleb:
It would be good to also update the README, as the change would add full support for groff ms AFAIU.

a-vrma:
Updated README and added a test.

tarleb:
Thanks!

PR #125, opened 2020-10-19
Add descriptions of each filter to the main README.md

the-solipsist:
This pull request helps address the problem of not knowing what's in each filter's folder in case the filter title isn't descriptive enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,16 +8,31 @@ This repository collects Lua filters for pandoc.
 To learn about Lua filters, see the
 [documentation](http://pandoc.org/lua-filters.html).
 
-Structure
----------
+Summary
+-------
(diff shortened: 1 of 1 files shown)


jgm:
This would require manual modification of the main README.md any time a filter is added.

We could adopt the convention that the first paragraph of a filter's README.md should be a short general description (this is true already in most if not all cases), and then create a Makefile target that updates the main README.md from these.

(Also, it would be good to keep line lengths to < 80 characters, as is the case throughout most of this repository.)

jgm:
Of course, another option would be to document modifying the main README.md as part of the filter submission process.

the-solipsist:
I feel adding to the main README.md as part of the filter submission process would be the way to go.  I feel having an overview of all the filters on a single page is very useful.

alerque:
I remember first finding and exploring this repository and wishing it had an overview index like this. The upkeep is a little bit of work, but I think making the filters more approachable is probably worth it.

PR #130, opened 2020-10-26
include-code-files: inclusion of code from source files (as pandoc-include-code hs filter)

b3:
In order to facilitate an eventual inclusion of code as commented in PR #126 I modified my fork and created distincts PR.

This one includes a new filter called `include-code-files` which mimic almost every feature of [pandoc-include-code](https://github.com/owickstrom/pandoc-include-code) which offer inclusion of source code files which some specific options (ranges, dedent).


include-code-files/include-code-files.lua
```diff
@@ -0,0 +1,51 @@
...
+  return line:sub(1,n):gsub(" ","") .. line:sub(n+1)
+end
+
+--- Filter function for code blocks
+local function transclude (cb)
+  if cb.attributes.include then
+    local content = ""
+    local fh = io.open(cb.attributes.include)
+    if not fh then
+      io.stderr:write("Cannot open file " .. cb.attributes.include .. " | Skipping includes\n")
```
tarleb, line 17: Would it make sense to just `return nil` from the function at this point? This would allow us to remove the `else` clause and thereby reduce nesting.

include-code-files/include-code-files.lua
```diff
@@ -0,0 +1,51 @@
+--- include-code-files.lua – filter to include code from source files
+---
+--- Copyright: © 2020 Bruno BEAUFILS
+--- License:   MIT – see LICENSE file for details
+
+--- Dedent a line
+local function dedent (line, n)
+  return line:sub(1,n):gsub(" ","") .. line:sub(n+1)
```
tarleb, line 8: I love this line. <3

tarleb (approved):
Thank you, this is great. I left a short inline comment, but I'll go ahead and merge now.

One possible enhancement would be to support hyphenated names like `start-line` as alternatives to the `startLine` pascalCase names. Pandoc uses hyphenated names in many places, so that might make this more natural to use.

include-code-files/README.md
```diff
@@ -0,0 +1,54 @@
+# include-code-files
+
+Filter to include code from source files.
+
+The filter is largely inspired by [pandoc-include-code](https://github.com/owickstrom/pandoc-include-code).
```
tarleb, line 5: Overlong line.

include-code-files/include-code-files.lua
b3: As far as `pandoc.CodeBlock(''')` returns `nil` and that no other filters use the attributes placed in the code this is of course a nice idea for simplifying the code.
As I said I, however, do not understand pandoc lua filter process enough to be confident when I wrote that piece of code.
Can't we imagine that another filter on `CodeBlock` may be use after this one? That was my assumption and the reason for cleaning specific attributes used only by this filter.

tarleb: I guess that's reasonable. I usually follow the principle that if an element cannot be processed the way we'd like, then it won't be touched at all. Whether or not that applies here is open to interpretation. I'm fine with the way it is now.

include-code-files/README.md
b3: Correction proposed in PR #136.

b3:
I created a PR #137 to **add** support for `start-line` and `end-line` alternatives attributes replying to your above comment.
I first used PascalCase versions because it is what is proposed in current pandoc version (as in `numberLines` or `startFrom` for instance).

tarleb:
Thanks. Good point about numberLines, I hadn't considered that.

include-code-files/README.md
b3: Fixed in PR #141 instead of #136 (I cleaned my fork)

PR #131, opened 2020-11-08
Update diagram-generator to work with pandoc-crossref

ctsa:
This change allows filters like pandoc-crossref and pandoc-fignos to reference the figure number created by diagram-generator.


tarleb:
Good idea, thanks!

piccolbo:
Did anyone notice a problem with this since pandoc v3? My refs to diagrams are now broken.

PR #132, opened 2020-11-10
diagram-generator: Asymptote graphics

jkseppan:
Implement an [Asymptote](https://asymptote.sourceforge.io/) filter in diagram-generator. Rename the `convert_from_pdf` function to `convert_with_inkscape` since it works just as well for svg source files.


tarleb (approved):
Thank you! We'll also need to install asymptote on our test environments. I can do that, as I'll have to redo the CI setup anyway, now that Travis CI's usefulness has finally fallen victim to their new owner's business strategy (see #134).
I am going to merge this once the tests work again.

PR #133, opened 2020-11-10
Replace paths in `include-files.lua` and filter blocks by format

gabyx:
Here is a PR with tests which provide some attributes to include and exclude include blocks as 
well as replace extensions depending on the output format. 

This comes in very handy for example when you want another file to be included for a different format. Application is a `table.md` with a HTML table which you format/style manually for Latex -> replace extensions with `.tex`. See the `README.md`.

Enhancements for filter: `include-files.lua`
- Support for attribute `replace-ext-if-format`.
- Global meta variable `include-format` to use on
  all include blocks.

New block filter `block-filter-format.lua`:
- Filtering blocks depending on attribute `include-if-format`.
- Filtering blocks depending on attribute `exclude-if-format`.

- Documentation in `README.md`.
- Tests added.

Would be cool if we could support this.


tarleb (commented):
This looks useful. In fact, I think there might be one or two orthogonal filters hiding in here. The only thing that's used from the existing code is the recognition of `include` markers on code blocks. Everything else appears to be independent and could be implemented as a filter that modifies the code blocks and is run before `include-files`.

gabyx:
Ah jeah. Definitely. Any suggestion, on the naming?

gabyx:
- The include/exclude part of an own filter.
- The filepath extension replacement would stay in this filter

- Bonus: Maybe include an environment replacement `replace-env-vars` if that is possible with Lua... (other PR :-))

gabyx:
I am going to call it `codeblocks-filter-format`.

gabyx:
See latest commit. Its split up.

gabyx:
Thanks for merging.

gabyx:
Ok. I now made it more generic: All blocks with attributes can be filtered now.

gabyx:
> ... Everything else appears to be independent and could be implemented as a filter that modifies the code blocks and is run before `include-files`.

No it cant: Its a chicken an egg problem: Every filter which runs before this filter will eventually only run on the main document.
(Chicken-egg problem). 

I close this in favor of a newer PR base on master.

alerque:
You don't need to close PR's to rebase them. You can just rebase and push to the branch this PR is opened against and it will update.

gabyx:
I know. Thought because of the mess, I make it more clear. 

Von meinem iPhone gesendet

PR #135, opened 2020-11-11
CI: use GitHub actions with Docker to run tests

tarleb:
Closes: #135


tarleb:
I believe I don't have the necessary permissions to disable Circle CI. @jgm, could you take a look and merge this when the other CIs are disabled?

jgm:
looks good!

PR #138, opened 2020-11-11
lilypond: avoid coreutils commands

tarleb:
Files are no longer created in the file system but only added to the
mediabag.  The files can still be created by running pandoc with the
`--extract-media` option. This allows to only depend on lilypond; no
other external commands are called.

The old behavior could be restored with a separate filter if desired.


alerque (approved):

tarleb:
@cole-miller could you take a look whether this somehow creates problems?

tarleb:
Merging now, but I'm open to rolling this back should any issues arise.

PR #149, opened 2020-12-02
Fix table-short-captions support for Pandoc 2.10+

smancill:
Support for Pandoc 2.10 is not working, and all the tests are broken too, they do not check anything.

This PR fixes Pandoc 2.10 support, and also fixes the tests to ensure they do run (and updates them to the changes introduced by  7a49118).

Commit 6e1b2cc introduced support for Pandoc 2.10 but it doesn't work because it treats the new `Table.caption.long` object as the pre 2.10 `Table.caption` object, but they are completely different lists. `Table.caption.long` is a list of `Blocks`, where each block has a `content` key which is a `List` of inlines (or it should be, a test would be needed for more complex captions?). The pre 2.10 `Table.caption` is a `List` of inlines, so they cannot share the same code to be manipulated.

This PR fixes it by just splitting the code into two different functions. They do follow the same structure, but it is better to keep them separated, otherwise it would require too many ifs, or code that is not clear.

For pre 2.10 `Table.caption` just keep the code as it is. For the new `Table.caption.long`, ensure the outer list of `Blocks` is taken into account when searching for the inline `Span` with the short caption.

Now all test do run and all pass, with pre and post Pandoc 2.10 (`nix-run-version` is a helper that allows  me to use [nixpkgs](https://github.com/NixOS/nixpkgs) releases to run different versions of a command)

Before this PR (well, not really, this PR also fixes the tests to see them failing. Before the last commit fixing 2.10 support):

``` shell
nix-run-version 19.09 pandoc make && echo "no errors"
pandoc 2.7.3
Compiled with pandoc-types 1.17.6, texmath 0.11.2.2, skylighting 0.8.2
no errors

$ nix-run-version 20.09 pandoc make && echo "no errors"
pandoc 2.10.1
Compiled with pandoc-types 1.21, texmath 0.12.0.2, skylighting 0.8.5
Output does not contain `\def\pandoctableshortcapt{}  % .unlisted`.
make: *** [test] Error 1
```

With this PR:

``` shell
$ nix-run-version 19.09 pandoc make && echo "no errors"
pandoc 2.7.3
Compiled with pandoc-types 1.17.6, texmath 0.11.2.2, skylighting 0.8.2
no errors

$ nix-run-version 20.09 pandoc make && echo "no errors"
pandoc 2.10.1
Compiled with pandoc-types 1.21, texmath 0.12.0.2, skylighting 0.8.5
no errors

$ nix-run-version unstable pandoc make && echo "no errors"
pandoc 2.11.2
Compiled with pandoc-types 1.22, texmath 0.12.0.3, skylighting 0.10.0.3,
citeproc 0.2, ipynb 0.1.0.1
no errors
```

### DIFF
--- a/table-short-captions/table-short-captions.lua
+++ b/table-short-captions/table-short-captions.lua
@@ -92,27 +92,24 @@ local function parse_table_attrs(attr)
   return label, short_caption, unlisted
 end
 
---- Wraps a table with shortcaption code
+function is_properties_span(inl)
+  return (inl.t) and (inl.t == "Span")                      -- is span
+                 and (inl.content) and (#inl.content == 0)  -- is empty span
+end
+
+--- Parse the caption for Pandoc < 2.10
 -- @tparam Table tbl : The table with {}-wrapped properties in the caption
--- @treturn List[Blocks] : The table with {label} in the caption,
---                         optionally wrapped in shortcaption code
-function rewrite_longtable_caption(tbl)
-  local caption
-  if PANDOC_VERSION >= {2,10} then
-    caption = pandoc.List(tbl.caption.long)
-  else
-    caption = tbl.caption
-  end
+-- @treturn ?string : The "short-caption" property, if present.
+-- @treturn bool : Whether ".unlisted" appeared in the classes
+function parse_short_caption_legacy(tbl)
+  local caption = tbl.caption
+
   -- Escape if there is no caption present.
   if not caption or #caption == 0 then
     return nil
   end
 
   -- Try find the properties block
-  local is_properties_span = function (inl)
-    return (inl.t) and (inl.t == "Span")                      -- is span
-                   and (inl.content) and (#inl.content == 0)  -- is empty span
-  end
   local propspan, idx = caption:find_if(is_properties_span)
 
   -- If we couldn't find properties, escape.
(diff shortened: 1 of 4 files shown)


smancill:
@blake-riley please review.

smancill:
Agh, it works with every Pandoc version I tried on my Mac. I need to figure out why the CI is failing.

smancill:
Rebased to fix the CI (`$(</dev/stdin)` was not working so I went back to `$(cat -)`).

table-short-captions/table-short-captions.lua
```diff
@@ -132,13 +129,69 @@ function rewrite_longtable_caption(tbl)
...
+  -- If we couldn't find properties, escape.
+  if not propspan then
+    return nil
+  end
+
+  -- Otherwise, parse it all
+  local label, short_caption, unlisted = parse_table_attrs(propspan.attr)
+
+  -- Excise the span from the caption
+  caption[bidx].content[idx] = nil
```
tarleb, line 169: Wouldn't this truncate the caption if `#caption[bidx] ~= idx`? How about
``` lua
caption[bidx].content[idx] = #caption[bidx] ~= idx and pandoc.Span{} or nil
```

Alternatively, we could also expect the properties span to always be the last element in the last block, which would then simplify `find_properties`.

tarleb (approved):
Looks great, thank you! I left one inline comment, but otherwise this is ready to be merged.

table-short-captions/table-short-captions.lua
smancill: I followed the original code, which uses

``` lua
  caption[idx] = nil
```

It would cause the same issue if `#caption ~= idx` in that case, but I didn't want to refactor the plugin too much.

I guess the properties span will always appear at the end (it is expecting the attributes to be converted to span by adding the `[]`, and the attributes always are at the end).

Maybe another PR to not mix too much changes in this one?

tarleb: I see, that's very reasonable.
A separate PR would be welcome!

PR #150, opened 2020-12-31
Adds filter: Columns

jdutant:
Adds Columns, a filter for multicolumn support in pandoc's markdown with html and LaTeX outputs.

### DIFF
--- a/columns/.gitattributes
+++ b/columns/.gitattributes
@@ -1,0 +1,2 @@
+# Auto detect text files and perform LF normalization
+* text=auto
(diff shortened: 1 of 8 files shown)


jdutant:
Thanks Albert, no worries: I am happy to leave that in its separate repo [https://github.com/jdutant/columns](https://github.com/jdutant/columns) and post it on the wiki instead (sorry I wasn't aware of the wiki actually). You can close the PR if that's easier.

Also, I plan to modify it by converting LaTeX longtables into xtabs within columns environments (as suggested in [this pandoc issue](https://github.com/jgm/pandoc/issues/1023)).

PR #151, opened 2021-01-06
Filter for linguistic examples

cysouw:
New Lua-filter to make examples according to the field of linguistics. Proposes a logical structure for such examples, a way to edit them in Pandoc-markdown, parses such examples, and preparing output that works well with html, epub, docx. Exporting to various Latex packages dealing with the same problem are included.


tarleb:
Thank you for the PR, this looks very impressive!

Linguistics is entirely outside my area of expertise, and I'm not sure I understand enough of what this does in details. We generally try to make sure that we can maintain all code in this repo ourselves in case the original author of a script goes missing. Given the massive size and unfamiliar topic of this filter, I'm uncertain whether we can achieve this. There are other Lua filters maintained as separate projects, like [pandoc-zotxt](https://github.com/odkr/pandoc-zotxt.lua). I feel like this might be a better option here. @jgm, what do you think?

I'll be happy to give feedback on the code either way. As a first comment, may I suggest to use [editorconfig](https://editorconfig.org)? It helps us to maintain a consistent coding style, and (mostly) prevents issues like mixing tabs and spaces.

cysouw:
The code is also available in my own repo: <https://github.com/cysouw/pandoc-ling>, so I am fine either way.

However, I find the idea of a centralised point for Lua Filters very attractive, because it makes it easer for users to install and use them. However, I understand the problem of maintenance!

cysouw:
BTW: @jgm is there any plan to include some of these filters in the regular Pandoc install?

alerque (changes requested):
The `.DS_Store` files included as part of this PR should be removed (by amending the existing commits that added them so they never show up it history at all). These files are specific to *your* macOS system and should never be tracked in any project's Git repository.

tarleb:
Suggestions to make the repo easier to find: 
- add the `pandoc` and `pandoc-filter` topic to your repo;
- enter the project into the list of [pandoc filters on the wiki](https://github.com/jgm/pandoc/wiki/Pandoc-Filters);
- announce it on [pandoc-discuss](https://groups.google.com/forum/#!forum/pandoc-discuss).

I still plan to setup a website for pandoc filters, and we could link it from there as well.

cysouw:
OK, then just close this PR here and I'll keep working on it separately. thanks for the interest and the kind words!

(and BTW: maybe put the .DS_Store in the gitignore of this Repo, then they wouldn't come through with PR from macOS-users)

tarleb:
> (and BTW: maybe put the .DS_Store in the gitignore of this Repo, then they wouldn't come through with PR from macOS-users)

It's usually better to let users handle the idiosyncrasies of their systems and editors, e.g. by encouraging them to exclude auto-generated files via a global `~/.gitignore`. But I don't feel strongly about it, so if you'd like to make a PR, I'd accept it.

If you are on Twitter and want to increase visibility there, then ping `@pandoc_tips` so we can retweet.

jgm:
I agree:  a big filter like this is probably better developed in a separate project.

> BTW: @jgm is there any plan to include some of these filters in the regular Pandoc install?

No.

PR #153, opened 2021-01-16
Provides not-in-format filter

jdutant:
A simple filter to keep parts of documents out of selected formats. Useful if you want to provide Raw code for some but not all output formats, e.g., if you have a markdown table that works well in formats other than LaTeX and you want to provide alternative raw code for LaTeX output only.

### DIFF
--- a/not-in-format/not-in-format.lua
+++ b/not-in-format/not-in-format.lua
@@ -1,0 +1,35 @@
+--- not-in-format - keep document parts out of selected output formats.
+--
+-- This Lua filter for Pandoc that keeps parts of a document out of
+-- selected outputs formats.
+--
+-- v1.0
+--
+-- @author Julien Dutant <julien.dutant@kcl.ac.uk>
+-- @copyright 2021 Julien Dutant
+-- @license MIT - see LICENSE file for details.
+--
+
+--- Test whether the target format is in a given list
+-- @param formats list of formats to be matched
+-- @return true if match, false otherwise
+local function format_matches(formats)
+  for _,format in pairs(formats) do
+    if FORMAT:match(format) then
+      return true
+    end
+  end
+  return false
+end
+
+--- Main Div filter
+-- Removes Divs tagged with `not-in-format` if the
+-- output format matches one of the Div's attributes
+function Div(element)
+  if element.classes:includes('not-in-format') and
+    format_matches(element.classes) then
+      return {}
+    else
+      return element
+    end
+end
--- a/not-in-format/not-in-format.lua
+++ b/not-in-format/not-in-format.lua
@@ -24,7 +24,7 @@ function Div(element)
   if element.classes:includes('not-in-format') and
     element.classes:find_if(function (x) return FORMAT:match(x) end) then
       return {}
-    else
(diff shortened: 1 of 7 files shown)


not-in-format/not-in-format.lua
```diff
@@ -0,0 +1,35 @@
...
+  end
+  return false
+end
+
+--- Main Div filter
+-- Removes Divs tagged with `not-in-format` if the
+-- output format matches one of the Div's attributes
+function Div(element)
+  if element.classes:includes('not-in-format') and
+    format_matches(element.classes) then
```
tarleb, line 31: Could be shortened to `element.classes:find_if(function (x) return FORMAT:match(x) end)`.

not-in-format/not-in-format.lua
```diff
@@ -0,0 +1,36 @@
...
+-- Removes Divs tagged with `not-in-format` if the
+-- output format matches one of the Div's attributes
+-- @param element a Div element
+-- @return an empty block if the Div classes include `not-in-format` and
+--  and the target format
+function Div(element)
+  if element.classes:includes('not-in-format') and
+    format_matches(element.classes) then
+      return {}
+    else
```
tarleb, line 33: Omitting the `else` clause would produce the same results, but is has slightly better performance: if a function returns nil, then the element is kept unchanged; if a value is returned then pandoc then has to parse it.
On the other hand, the current version is probably easier to understand, so no change required.

tarleb (approved):
Thanks, I like this very much!
Please just make sure that all lines are kept at 80 chars or below, including in Markdown files.

not-in-format/not-in-format.lua
```diff
@@ -1,14 +1,12 @@
...
 -- This Lua filter for Pandoc that keeps parts of a document out of
 -- selected outputs formats.
 --
--- v1.0
---
 -- @author Julien Dutant <julien.dutant@kcl.ac.uk>
 -- @copyright 2021 Julien Dutant
 -- @license MIT - see LICENSE file for details.
---
+-- @release 1.0
```
tarleb, line 9: This is a good idea, we should add such a line to all filters

not-in-format/not-in-format.lua
jdutant: Elegant, thanks. I've simplified the code - but added a comment to explain what it does, it wasn't immediately obvious to me and it might help other readers.

not-in-format/not-in-format.lua
jdutant: I see, I didn't know that. Better make it more efficient and clarify in the comments.

not-in-format/not-in-format.lua
jdutant: I'm planning to use [LDoc documentation format](https://stevedonovan.github.io/ldoc/) in lua filters. Perhaps something to (gently) suggest to potential contributors?

jdutant:
I've implemented the suggestions (and credit). It's basically a one-line filter now, but should be of use.

tarleb:
Thanks!

PR #160, opened 2021-02-08
multiple-bibliographies: Set the div attributes and classes according to the attributes of the cs:bibliography element

badumont:
In the CSL stylesheet, the element `cs:bibliography` may carry various attributes determining the rendering of the bibliography that citeproc converts into classes and attributes for the references div. As the filter simply puts the partial bibliographies created by citeproc into the custom divs, these classes and attributes are not set (excepted `csl-bib-body`). I propose to store the div created by citeproc in order to copy its classes and attributes to the custom divs. 

It has the side effect that all classes and attributes manually set by the user for the corresponding divs will be ignored. I think it more as a feature than as a drawback, since one probably should not do this anyway, but modify the CSL file instead. However, if the possibility of overriding the CSL file is desired (to use different layouts across the document?), the proposed modification can be expanded so that each class or attribute of `refs_div_with_properties` is added to `div` only if it is not already set.


tarleb:
Excellent, thanks!

PR #161, opened 2021-02-12
Adds ending abstracts with a horizontal rule

jdutant:
To place an abstract at the beginning of a paper that doesn't start with a heading.


tarleb:
I manually cherry-picked the commit and fixed a type and line lengths. It's now on the master branch. Thanks!

PR #163, opened 2021-02-13
Bibliography place: control the placement of a citeproc bibliography in a Pandoc template

jdutant:
Useful to place something between the end of text and the bibliography when using citeproc. The filter places any citeproc generated bibliography Div in a metadata variable, `referencesblock`, that can then be printed anywhere in a custom template. Since the entire Div is passed on, any formating pandoc/citeproc applies to it for various outputs is preserved.

The filter does not affect biblatex / natbib outputs, so it can be used with a custom template that caters for all bibliography engines.


jdutant:
In the first version only the Div with identifier`refs` was captured. But this may be preceded by a header. In the new version the filter captures both the references and whatever heading immediately precedes them.

PR #169, opened 2021-03-21
First line indent

jdutant:
Not sure whether this filter is simple enough for this repo. 

The filter provides finer control of the first-line indentation style of separating paragraphs in Pandoc's markdown, targetting HTML and LaTeX/PDF outputs. It switches on Pandoc's first-line
indent style, automatically removes (typically unwanted) first-line indents after blockquotes and
lists unless the user specifies otherwise, and allows users to control whether a paragraph has first-line indentation in both LaTeX and HTML by inserting `\indent` or `\noindent` at the beginning of a paragraph. 

Limitation: at present it only handles English-style typography in HTML (in French and perhaps other languages, paragraphs are first-line indented even after headings; this is handled in LaTeX by the babel or polyglossia languages, not re-implemented for HTML in that filter.)


tarleb:
Looking great, thanks Julien!

PR #172, opened 2021-04-19
Set working directory during recursion

lmunch:
This change will prepend the working directory of the included file
to the Image elements src path. E.g if file included is
subdoc/somefile.md then subdoc is preprended to any image element
in somefile.md

This change also fixes include files with relative names when
nesting more than one level.


tarleb:
Thanks!

PR #173, opened 2021-04-24
Include files filter updates

lmunch:
This patchset keeps the paths to images relative and also improves the usage in combination with the include-code-files.lua filter

### DIFF
--- a/include-files/Makefile
+++ b/include-files/Makefile
@@ -10,6 +10,6 @@ expected.native: sample.md file-a.md file-b.md file-c.md include-files.lua
 	pandoc --lua-filter=include-files.lua --output $@ $<
 
 expected-auto.native: sample.md file-a.md file-b.md file-c.md include-files.lua
-	pandoc --lua-filter=include-files.lua --output $@ $<
+	pandoc --lua-filter=include-files.lua -M include-auto --output $@ $<
 
 .PHONY: test
--- a/include-files/README.md
+++ b/include-files/README.md
@@ -62,10 +62,13 @@ Only plain-text formats are accepted.
 ### Recursive transclusion
 
 Included files can in turn include other files. Note that all
-filenames must be relative to the directory from which pandoc is
-run. I.e., if a file `a/b.md` is included in the main document,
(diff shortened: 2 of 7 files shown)


lmunch:
@tarleb Please review

tarleb:
Thanks, looks good! Can we add a test for this?

lmunch:
Sure, I can add a test. I will look at it later this week.

lmunch:
@tarleb I have updated the test and documentation, but the CI fails in math2svg which seems to be unreleated.

BTW, awesome work with pandoc, pandoc/latex docker image etc. Use it everyday, highly appreciated.

tarleb:
Awesome, thanks! I'll look into the math2svg problems.

tarleb:
It seems that the math2svg problems are caused by changes in the npm ecosystem. It's probably enough to configure a Ubuntu PPA to get a newer node version installed in our testing container. I'll try if that fixes it, and just merge regardless if it doesn't.

gabyx:
@lmunch : The relative path changes is generally a good idea, but its really bit too intrusive because

- generally one never uses **absolute** paths in markdown documents, so one cannot have include paths relative to the **working directory anymore** which is sad (since images might be inside `files/img.png` which is relative to the current working dir.

This feature needs probably more thoughts:

A variant with absolute paths like:

````md
```{.include}
${env:ROOT_DIR}/chapters/A.md
```
![](${env:ROOT_DIR}/files/img.png)
````

will work if `${env:ROOT_DIR}` is replaced beforehand with the environment variable `ROOT_DIR` which can points to the root of the document tree.

In this way one can still write multimarkdown documents.

Or maybe 

````md
```{.include .include-is-relative}
chapters/A.md
```
![](files/imgA.png){.include-is-relative}
![](files/imgGeneral.png)
````

I would preferably do the above, only change paths if `.include-is-relative` is used.
On can then also introduce a meta variable `includes-are-relative=true` which will default to your version, which is intrusive.

gabyx:
I added this feature in #177.
See the Readme under transclusion.

PR #174, opened 2021-05-01
Add complex-paragraphs

badumont:
This filter prevents the indentation of Para objects wrapped in a Div with class `.complex-paragraph` (except the first one). It is useful, for example, when a paragraph contains a quotation.


tarleb:
I'm sorry, I had forgotten about this. Would you still like us to review it?

badumont:
Sorry, I just noticed that I had not created a branch and I that it would cause problems. I am opening another one!

PR #178, opened 2021-05-28
Add filter doi2cite

korintje:
Add a filter `doi2cite`, which helps users to insert references into a document with using DOI(Digital Object Identifier) tags.

# What the filter do
With this filter, users do not need to make bibtex file by themselves. 
Instead, the filter automatically generate .bib file from the DOI tags (e.g. [@DOI:xxxx.xxxx.xxx.]), and convert the DOI tags into citation keys available by pandoc --citeproc (e.g. [@Korintje_2021]).

# Filter features
- Get bibtex data of the DOI from http://api.crossref.org
- Can be mixedly used  with .bib files manually collected by users
- Use saved bibtex data from the second time to avoid unnecessarily network accesses
- Detect and avoid duplications of bibtex keys.

I'd be glad if I can get any comments or suggestions!


tarleb:
Thank you, this looks very useful! It's going to be about a week before I have time to review this in detail, but maybe one of the other maintainers can do so earlier.

alerque:
I'm not really qualified to review this; I'm not familiar with the assorted citation ecosystems or any of the issues surrounding them. It does look like you've done good work getting it setup, I'm just vaguely aware there are a lot of pitfalls/complexities in this area and don't have the eye to spot the or any test content to run it on.

gnpan:
No problems so far. Also works well with author-date citation formats, respecting author suppression with [-@doi...] or citations without brackets to obtain Author et al. (2021) instead of (Author et al., 2021). If you want to use a second lua filter it's best to first create the expected.md and then convert that to pdf with citeproc etc.

korintje:
Thank you for the reviews and suggestions!
- I fixed over long lines in README (excepting long URL lines)
- I changed indent spaces in doi2cite.lua from 4 to 2.

It would be smarter to substitute `pandoc.path` for `split_path`, but in my environment (pandoc 2.14.0.1 @ WSL Ubuntu 20.04), the field `pandoc.path` is nil, so I left it as it is for now.

tarleb:
Great, thanks!

>  but in my environment (pandoc 2.14.0.1 @ WSL Ubuntu 20.04), the field `pandoc.path` is nil,

Oh, that might be worth fixing in pandoc. You could still get it by using `pandoc.path = require 'pandoc.path'`.

korintje:
Fixed an error in the CI test.
I think the error occurred because `test` process tried to get a differences between two binary .pdf files.
I rewrote it to take the difference between "expected.md" and "output.md", mimicking examples of other filters.

tarleb:
Thanks again!

PR #181, opened 2021-06-21
support for mermaidjs [enhancement]

maehr:
### Mermaid

To use Mermaid, you must install mermaid-cli itself. See the
[mermaid-cli website](https://github.com/mermaid-js/mermaid-cli) for more details.

This filter assumes that the `mmdc` command is located in the path
and therefore can be used from any location. Alternatively, you can
set the environment variable `MERMAID` or use the pandoc's meta variable
`mermaid_path`.

Example usage from [the Mermaid page](https://github.com/mermaid-js/mermaid):

```{.mermaid caption="This is a sequence diagram, created by **Mermaid**."}
sequenceDiagram
Alice->>John: Hello John, how are you?
loop Healthcheck
    John->>John: Fight against hypochondria
end
Note right of John: Rational thoughts!
John-->>Alice: Great!
John->>Bob: How about you?
Bob-->>John: Jolly good!
```

---

A possible fix for https://github.com/pandoc/lua-filters/issues/93

### DIFF
--- a/.tools/Dockerfile
+++ b/.tools/Dockerfile
@@ -15,13 +15,49 @@ RUN apt-get -q --no-allow-insecure-repositories update \
        aspell-en \
        aspell-fr \
        asymptote \
+       ca-certificates \
+       chromium-browser \
        default-jre \
        dvisvgm \
+       fonts-liberation \
        graphviz \
        imagemagick \
        inkscape \
+       libappindicator3-1 \
+       libasound2 \
+       libatk-bridge2.0-0 \
+       libatk1.0-0 \
+       libc6 \
+       libcairo2 \
+       libcups2 \
+       libdbus-1-3 \
+       libexpat1 \
+       libfontconfig1 \
+       libgbm1 \
+       libgcc1 \
+       libglib2.0-0 \
+       libgtk-3-0 \
+       libnspr4 \
+       libnss3 \
+       libpango-1.0-0 \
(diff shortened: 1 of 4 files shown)


tarleb:
Thanks!

I tried to install mermaid-cli in the Docker image which we use for testing, but failed due to some phantomjs problems. I'm not a JS or npm expert, so not sure yet what's happening there.

maehr:
> Thanks!
> 
> I tried to install mermaid-cli in the Docker image which we use for testing, but failed due to some phantomjs problems. I'm not a JS or npm expert, so not sure yet what's happening there.

I did some debugging and puppeteer doesn't like running as root... <https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md>

It's hard to fix without breaking everything else.

maehr:
I added some dependencies according to <https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#chrome-headless-doesnt-launch-on-unix> and turned the npm flag `--unsafe-perm` on. Now it passes. Tested with <https://github.com/nektos/act>.

maehr:
@garfvl thank you very much for fixing it. I failed a year ago. Please go ahead. If I can help in any way let me know.

tarleb:
The diagram generator is now maintained at https://github.com/pandoc-ext/diagram; I've added Mermaid support there.

Thank you for your contribution!

PR #184, opened 2021-07-04
Fixes bug when header-includes is a list

jdutant:
When header-includes is a list the filter crashes. This PR fixes this.


jdutant:
Needs a further fix

PR #185, opened 2021-07-04
First-line-indent: fix bug with pre-existing header-includes

jdutant:
The first-line-indent filter crashed when the original document had header-includes already. This is now fixed, tested for all types of header-includes (MetaInlines, MetaBlocks and MetaList).

### DIFF
--- a/first-line-indent/first-line-indent.lua
+++ b/first-line-indent/first-line-indent.lua
@@ -82,10 +82,10 @@ local header_code = {
 -- @return meta the modified metadata block
 local function add_header_includes(meta, blocks)
 
-  local header_includes = pandoc.List(blocks)
+  local header_includes = pandoc.MetaList( { pandoc.MetaBlocks(blocks) })
 
   -- add any exisiting meta['header-includes']
-  -- it could be a MetaList or a single String
+  -- it can be MetaInlines, MetaBlocks or MetaList
   if meta['header-includes'] then
     if meta['header-includes'].t == 'MetaList' then
       header_includes:extend(meta['header-includes'])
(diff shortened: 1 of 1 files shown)


alerque:
Did this problem crop up because of Pandoc's (several) recent changes to header include handling or has this always been a problem?

alerque (approved):

jdutant:
Thanks. I'm not sure; I thought I had tested it on docs with header-includes and it worked. But it's also possible I overlooked it. Especially header-includes can have three types: MetaInlines, MetaBlocks and MetaList, and I may have failed to try it with all of them. I've now tested the new version with all. I haven't included sample .md files for all cases but could do if that's helpful for the future.

alerque:
Thanks for the response, and that's fair.

Actually I think including test fixtures would be quite helpful in this case.

bpj:
What I usually do is to check if `header-includes` is `nil`, a MetaList or
something else. If it is a MetaList I just push my additions onto it as a
MetaBlocks containing a RawBlock; otherwise I replace it with a MetaList
and push the original value if any and my own …Block onto it. Has never
failed so far (unless the --include-in-header option has been used.)

Just my two worthless coins.

Den ons 7 juli 2021 17:53Julien Dutant ***@***.***> skrev:

jdutant:
@bpj Thanks. That's right; this filter does the same, except that it first creates a MetaList with the desired content as MetaBlocks containing a Rawblock, then check if there's something in the original doc's Meta; if yes, then either it's a MetaList and we extend ours with it, or it's something else and we push it in ours. 

@alerque I've added three sample docs covering the different types of header-includes.

PR #191, opened 2021-10-23
A lua filter enabling multiple columns in Latex, PDF, and HTML documents

chrisaga:
It's based on the use of simple Pandoc fenced divs, their class names and some arguments. This filter extends what is already available for HTML output, gives the same features to Latex (and PDF via Latex) documents.
Multiple columns are very useful to format fancy documents with text, pictograms etc. I have been missing this feature for a long time. I finally made my first Pandoc lua filter and share it hoping it will be useful to others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,18 @@
 [GitHub Actions badge]: https://img.shields.io/github/workflow/status/pandoc/lua-filters/CI?logo=github
 [GitHub Actions]: https://github.com/pandoc/lua-filters/actions
 
+--------------------------------------------------------
+
+**DISCLAIMER : This repositoy is not active any more.**
+
+**See [hk-pandoc-filters](https://github.com/chrisaga/hk-pandoc-filters)**
+
+**I started this fork almost when pandoc/lus-filters stoped accepting third party contributions.**
+**Hence I started a fresh new repository to host my  own lua filters.**
+**As soon as they'll provide a way to contribute or be referenced, I will apply.**
+
+--------------------------------------------------------
+
 A collection of Lua filters for pandoc.
 
 To learn about Lua filters, see the [documentation].


column-div/column-div.lua
```diff
@@ -0,0 +1,115 @@
...
+Note:       You need to include multicol latex package to get balanced columns
+            in latex or pdf
+            I tried to use well known html or latex parameter.
+            Even if lua doen't like hyphens like in column-count.
+--]]
+local List = require 'pandoc.List'
+
+function Div(div)
+  options = ''
+  local env = div.classes[1]
```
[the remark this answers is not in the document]
alerque: I just checked through the code and still see several instances of `div.classes[1]` being used for logic. As @bpj outlined reasons for above this is a very brittle way to code this up. Yes the code is simpler and yes the Markdown might look cleaner too, but it is *magic* that will break as soon as somebody that thinks differently than you tries to use this following the usual rules of the game (i.e. attribute order has no semantic meaning). 

I consider this a blocker. This looks like it could be useful, but I don't want to see super brittle coding in here that will be a support and/or maintenance burden.

column-div/column-div.lua
```diff
@@ -0,0 +1,162 @@
...
+  -- TODO: use a div with no class to build a 'scope' in Latex
+  --        usefull for user who would throw inline Latex code and limit it's
+  --        effect.
+  if not div.classes or #div.classes == 0 then return nil end
+
+  -- if the output is beamer do columns
+  if FORMAT:match 'beamer' then
+    -- only arbitrary environment support in beamer for now. Environment has to
+    -- be the firs class name.
+    env = div.classes[1]
```
alerque, line 47: Assumptions like class name order make this too brittle to include in this repository. If it is made a bit more robust we could reconsider.

alerque (changes requested):

column-div/column-div.lua
```diff
@@ -0,0 +1,155 @@
...
+
+            See README.md for details
+
+Note:       You need to include multicol latex package to get balanced columns
+            in latex or pdf
+            I tried to use well known html or latex parameter.
+            Even if lua doen't like hyphens like in column-count.
+
+Bugs:       * html rendering throws a warning [WARNING] Ignoring duplicate attribute style="width:60%;".
+            when widht AND color are set and totally ignore the width
```
alerque, line 27:
```suggestion
            when width AND color are set and totally ignore the width
```

Lets figure out whether this bug is in this code or not. It should be pretty easy to setup an experiment to see if plain Pandoc with no filters has this problem, no?

column-div/expected.html
```diff
@@ -0,0 +1,216 @@
...
+    code{white-space: pre-wrap;}
+    span.smallcaps{font-variant: small-caps;}
+    span.underline{text-decoration: underline;}
+    div.column{display: inline-block; vertical-align: top; width: 50%;}
+    div.hanging-indent{margin-left: 1.5em; text-indent: -1.5em;}
+    ul.task-list{list-style: none;}
+    .display.math{display: block; text-align: center; margin: 0.5rem auto;}
+  </style>
+  <!--[if lt IE 9]>
+    <script src="//cdnjs.cloudflare.com/ajax/libs/html5shiv/3.7.3/html5shiv-printshiv.min.js"></script>
```
alerque, line 149: I am uncomfortable with this. Do we have other instances in Pandoc templates where the default output includes loading 3rd party external resources?

column-div/column-div.lua
```diff
@@ -0,0 +1,155 @@
...
+    if opt then
+      -- add color to style
+      style = 'background-color: ' .. opt .. ';' .. (style or '')
+      div.attributes['background-color'] = nil    -- consume attribute
+    end
+    -- if we have style then build returned list
+    if style then
+      div.attributes.style = style .. (div.attributes.style or '')
+      returned_list = List:new{pandoc.Div(div.content, div.attr)}
+      --returned_list = List:new{pandoc.Div(div.content)}
```
alerque, line 151: I realize I'm reviewing this still in Draft mode so it might be too early, but just as a note for when this *is* ready for review lets cleanup the code so there are not instances of commented out bits without explanation for why the logic is not transparent.

column-div/column-div.lua
```diff
@@ -0,0 +1,155 @@
...
+      returned_list = begin_env .. div.content .. end_env
+
+    else
+      -- other environments ex: multicols
+      if div.classes:includes('multicols') then
+        env = 'multicols'
+        -- process supported options
+        opt = div.attributes['column-count']
+        if opt then options = '{' .. opt .. '}' end
+      --[[ This functionality will be moved in another filter since it can't be consistent with the positionless classname requirement
```
alerque, line 104: c.f. Comment on L151.

alerque (changes requested):
Thanks for working on this, dropping the requirement on specific class order, and especially for adding tests!

column-div/column-div.lua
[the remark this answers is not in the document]
alerque: Latest commits seem to have this suggested fix..

column-div/expected.html
chrisaga: Well … I didn't do this. `pandoc --standalone` did. I can remove the `--standalone` option on the html test file.
I already did for the latex test file.
Would it be better like this ?

column-div/column-div.lua
chrisaga: Of course !

column-div/column-div.lua
chrisaga: 1) Plain Pandoc is OK
2) Pandoc + filter **seems to be** OK if the div uses only a `width` parameter (the filter **do no** process the `width` parameter)
3) Pandoc + filter **not** OK if there is a `width` parameter **and** at least a second parameter.

Two options so far :
1) I do something wrong in the code
2) I trigger a tricky bug

Need some time to figure out since it is my first Pandoc filter ever.

chrisaga: My best guess for now is that Pandoc doesn't check that a filter already set the `style` attribute so it attempts to create a new one with the value "width:xx%;".
I added the processing of the width attribute in the filter in addition to the other attributes. When the filter has nothing to do, it lets Pandoc process the width.

PR #202, opened 2021-12-21
Add filter list-table

not-my-profile:
Inspired by reStructuredText and the feature request https://github.com/jgm/pandoc/issues/7305.

### DIFF
--- a/list-table/Makefile
+++ b/list-table/Makefile
@@ -1,0 +1,8 @@
+DIFF ?= diff --strip-trailing-cr -u
+PANDOC ?= pandoc
+
+test: sample.md expected.html list-table.lua
+	@$(PANDOC) --lua-filter list-table.lua --to=html $< \
+		| $(DIFF) expected.html -
+
+.PHONY: test
--- a/list-table/README.md
+++ b/list-table/README.md
@@ -1,0 +1,47 @@
+# list-table
+
+reStructuredText has so-called [list tables], which are not
+only easy-to-write but also produce clean diffs since you
+don't need to re-align all the whitespace when one cell width
+changes. This filter lets you use RST-inspired list tables in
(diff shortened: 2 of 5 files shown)


mustafa0x:
Thanks for this! Is there any way to mark added fields?

I tried something like this but couldn't figure out how to fix it.

```lua
    for i = 1, #list.content do
        assert(#list.content[i] == 1, "expected item to contain only one block")
        assert(list.content[i][1].t == "BulletList",
               "expected bullet list, found " .. list.content[i][1].t)
        table.insert(rows, list.content[i][1].content)
    
        local rows1 = list.content[i][1]
        for j = 1, #rows1 do
          local row = rows1[j].content
          while #row < col_count do
            row_copy = {content={}}
            row_copy[1].content[1].text = 'ADDED_FIELD'
            row[#row + 1] = row_copy
          end
        end
    end
```

not-my-profile:
You're welcome :) I am afraid I don't understand what you're trying to achieve. What are "added fields"? Why do you want to mark them?

mustafa0x:
Sorry, I should've explained more!

```
$> pandoc --lua-filter list-table.lua

::: list-table
*   - a
    - b
*   - c
:::
```

```html
<table>
  <caption></caption>
  <thead>
    <tr class="header">
      <th>a</th>
      <th>b</th>
    </tr>
  </thead>
  <tbody>
    <tr class="odd">
      <td>c</td>
      <td></td> <!-- <<< added-->
    </tr>
  </tbody>
</table>
```

The optimal solution would be to add colspan=2, but I don't think this filter supports that, and it would require some work I believe. So as a workaround, if any added fields could be marked somehow, so I could later remove using `.replace('<td>MAGIC_STRING</td>', '')`.

Thank you!

PR #204, opened 2021-12-24
Update list-table

not-my-profile:
The new version introduces support for:

* headless tables
* tables with multiple header rows
* tables with header columns
* cell attributes (including colspan, rowspan and cell-specific alignment)

Furthermore the documentation has been improved.

Please don't squash my commits ;)

### DIFF
--- a/list-table/Makefile
+++ b/list-table/Makefile
@@ -1,8 +1,11 @@
 DIFF ?= diff --strip-trailing-cr -u
 PANDOC ?= pandoc
+CMD = $(PANDOC) --lua-filter list-table.lua --to=html sample.md
 
-test: sample.md expected.html list-table.lua
-	@$(PANDOC) --lua-filter list-table.lua --to=html $< \
-		| $(DIFF) expected.html -
+test:
+	@$(CMD) | $(DIFF) expected.html -
 
-.PHONY: test
+update:
+	$(CMD) > expected.html
+
+.PHONY: test update
--- a/list-table/README.md
+++ b/list-table/README.md
@@ -186,3 +186,47 @@ results in:
 </tr>
 </tbody>
 </table>
+
+## Cell attributes
+
+If the first inline element of a table cell is an empty Span it
+is removed and its attributes are transferred to the table cell.
+The `colspan` and `rowspan` attributes are supported.
+
+```
+:::{.list-table}
+   * - Name
+     - Italic
+     - Code
+
+   * - Markdown
+     - []{rowspan=2} `*italic*`
+     - `` `code` ``
+
+   * - reStructuredText
+     - ` ``code`` `
+:::
+```
+
+results in:
+
+<table>
+<thead>
+<tr>
+<th>Name</th>
+<th>Italic</th>
+<th>Code</th>
+</tr>
+</thead>
+<tbody>
+<tr>
+<td>Markdown</td>
+<td rowspan="2"> <code>*italic*</code></td>
+<td><code>`code`</code></td>
+</tr>
+<tr>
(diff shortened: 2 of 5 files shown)


mustafa0x:
Well, this is simply incredible.

tarleb:
This is great, thank you!

A heads-up: we are in the process of making tables a bit more user-friendly. E.g., we now have types and constructors `pandoc.TableHead`, `pandoc.TableFoot`, and `pandoc.Row` instead of using Attr-value pairs. This breaks backwards compatibility. See https://github.com/jgm/pandoc/issues/7718 for some context. I'm open to add limited backwards compatibility so your filter won't break, but you'd probably want to update some code either way.

Since you put significant time into working with tables, it would be great to hear your ideas on how to make the Lua interface better. Please feel free to open new issues if anything comes to mind.

not-my-profile:
Hey :) I don't mind updating my filter when the new Pandoc version with these constructors is released, so you don't have to add backwards compatibility just for my filter.

> Since you put significant time into working with tables, it would be great to hear your ideas on how to make the Lua interface better.

I noticed something about `SimpleTable`, which I reported yesterday (https://github.com/jgm/pandoc/issues/7776).

I am actually not very familiar with Lua so I don't think I can say much about the API design.

I think the formatting of the [Lua API docs](https://pandoc.org/lua-filters.html) could be improved by making it more vertically compact (i.e. displaying parameter name and description next to each other). `document.body.scrollHeight` reports 75987px for me which I think translates to 20 meters ... that's a bit long. With all of this up and down scrolling it's easy to get lost. Apart from making the page more compact, adding some table of contents that shows your current position would also help.

image

mustafa0x:
Minor: this failed to produce any output on pandoc 2.14.2 (no error though). Upgrading to latest (2.16.2) fixed.

```
$>pandoc --lua-filter list-table.lua --verbose

:::list-table
   * - row 1, column 1
     - row 1, column 2
     - row 1, column 3

   * - row 2, column 1
     -
     - row 2, column 3

   * - row 3, column 1
     - row 3, column 2
:::



[INFO] Running filter pandoc-list-table.lua
[INFO] Completed filter pandoc-list-table.lua in 6 ms
```

tarleb:
Curious. That usually happens when pandoc failed to retrieve the Table
object. You could try to wrap the Table in a list to get different
behavior and hopefully an error message.

We did a complete overhaul of the Lua subsystem in pandoc 2.15 and 2.16,
in part to avoid these kind of cryptic problems. Error reporting is much
better now – but still has weak points and needs a bit more work.

mustafa0x:
Another minor issue: `colspan` doesn't work with headers unless a redundant field is added. Worse, it'll remove later fields (see how no "cell 2" in output).

```
::: list-table
*  - []{colspan=2}header
*  - cell 1
   - cell 2
:::
```

```html
<table>
  <thead>
    <tr class="header">
      <th>header</th>
    </tr>
  </thead>
  <tbody>
    <tr class="odd">
      <td>cell 1</td>
    </tr>
  </tbody>
</table>
```

With redundant header field:

```
::: list-table
*  - []{colspan=2}header
   - IGNORED
*  - cell 1
   - cell 2
:::
```

```html
<table>
  <thead>
    <tr class="header">
      <th colspan="2">header</th>
    </tr>
  </thead>
  <tbody>
    <tr class="odd">
      <td>cell 1</td>
      <td>cell 2</td>
    </tr>
  </tbody>
</table>
```

This problem doesn't occur when using colspan for later rows.

not-my-profile:
@mustafa0x ah yeah, my bad ... will be fixed by #205.

PR #206, opened 2021-12-27
list-table: ensure compatibility with upcoming pandoc versions

tarleb:


tarleb:
@not-my-profile please let me know in case this doesn't look good to you.

not-my-profile:
LGTM, thanks! I didn't use the new constructors because I didn't know about them since aren't yet documented on https://pandoc.org/lua-filters.html. (btw why is that the case?)

tarleb:
Great, thanks!

The constructors are really new and only available in the nightly (which, BTW, will not work with the updated filter as `PANDOC_VERSION` is still set to 2.16.2). The docs will be updated once the new version is released.

Issue #207, opened 2021-12-31
Change this repository into a collection of links?


jgm:
One drawback of the current structure is that people submit code here but then don't monitor the repository, and issues are neglected.  Perhaps it would be better to make this simply a collection of links to lua filters that are maintained in independent repositories?

cagix:
I kind of like this idea. Maybe this repo could serve as a kind of collection of "official" scripts from the Pandoc creators and all other filters could be linked in the README (sorted by topic)? That would reduce the maintenance to checking the links every year. In addition, the "official" code provided could serve as a live demo / live documentation of the Lua API.

tarleb:
I'm *very* much in favor of that; it would save me a lot of time. It takes a significant amount of effort, on almost each new pandoc release, to adjust the tests and filters to the changes. It's tiresome, and pinging all authors and waiting for them to change the code would take just as long. I'd be glad to get out of that obligation.

We could still do occasional automatic "releases", which pack the filters into a single archive. This shouldn't be too hard if the individual repos use a common structure.

cagix:
> We could still do occasional automatic "releases", which pack the filters into a single archive. 

That sounds interesting, but might not be quite easy with regard to the then presumably different licences in different repos?

> This shouldn't be too hard if the individual repos use a common structure.

Would maintaining a template repo help with this?

tarleb:
Pinging everyone who contributed a filter so far: what do you think of this idea? What would you need to make this as painless as possible for you?

@jdutant @tolot27 @blake-riley @not-my-profile @svenevs @b3 @jkr @cole-miller @sokotim @korintje @gtuckerkellogg

cagix:
Since each filter belongs to a subfolder, it should be easy to split your repository into several individual repositories using `git filter-branch` and retain the individual history :)

not-my-profile:
I am not sure about this. There are advantages to having a common repository. Having them all here guarantees that other pepole can contribute improvements even when the original author has ceased maintenance.

>It takes a significant amount of effort, on almost each new pandoc release, to adjust the tests and filters to the changes. It's tiresome, and pinging all authors and waiting for them to change the code would take just as long. I'd be glad to get out of that obligation.

[...]

PR #208, opened 2022-01-04
Create math-katex filter

benabel:
Render math with katex server-side.


tarleb:
Thank you for this! Looks like a very useful filter, and the code is beautifully clean, too.

We're currently discussing the future path of this repo, and I'd like to wait for the discussion to reach a conclusion before merging. Please feel free to add your voice there, too: #207

math-katex/Makefile
```diff
@@ -0,0 +1,20 @@
+DIFF ?= diff --strip-trailing-cr -u
+PANDOC ?= pandoc
+
+test: sample.md math-katex.lua expected.html
+	@$(PANDOC) \
```
tarleb, line 5: Makefiles are weird in that they *require* the first level indentation to be with tabs, not space.

benabel: It is already with tabs on my editor.

tarleb: Ah, OK, great. I probably viewed a previous version, or my browser/github didn't show the file correctly.

tarleb: Indeed, looks like GitHub renders the tabs with a width of a single space. I didn't know that.

timtroendle:
@benabel, this looks very useful to me. Is your filter available somewhere?

benabel:
@timtroendle in fact I never realized it you can find it on my fork https://github.com/benabel/lua-filters/tree/master/math-katex.

PR #216, opened 2022-01-22
Tables vrules : adds vertical rules to tables (latex/pdf output) 

chrisaga:
Works with all kind of Pandoc supported Markdown tables.
- The Latex hack based on patching \LT@array doesn't work anymore for Multiline and Pipes tables with current versions of Pandoc.
- A Lua filter should hopefully have less compatibility issues and be easier to maintain.


tarleb:
Thanks for this. We put merging of new filters on hold for the time being, see #207 for details.

If you want to publish the filter in its own repository, please consider using https://github.com/tarleb/lua-filter-template and let us know about any problems with that repo.

PR #224, opened 2022-04-18
Add commonmark-code-blocks filter for producing CommonMark-suggested HTML for fenced code blocks

dmbaturin:
Pandoc's implementation of CommonMark code block rendering differs from that suggested by the [spec](https://spec.commonmark.org/0.30/#example-141) and used by most other Markdown convertors.

The suggested behavior is to output `<code class="language-$lang">`, while Pandoc omits the `language-` prefix, which may cause problems for syntax highlighters that rely on that spec-suggested behavior.

I made a simple filter to bring the code block rendering in line with the spec and other implementations.

It's used in one of my project's website build process for example, where pandoc is used as a [Markdown to HTML convertor](https://github.com/PataphysicalSociety/soupault.app/blob/master/soupault.toml#L23-L24), but syntax highlighting is [delegated to an external tool](https://github.com/PataphysicalSociety/soupault.app/blob/master/soupault.toml#L176-L181), and `*[class^="language-"]` CSS selector is used to find code elements with language specified.

### DIFF
--- a/commonmark-code-blocks/Makefile
+++ b/commonmark-code-blocks/Makefile
@@ -1,0 +1,11 @@
+DIFF ?= diff --ignore-all-space --ignore-blank-lines -u
+PANDOC ?= pandoc
+
+.PHONY: test
+test:
+	@$(PANDOC) --lua-filter=commonmark-code-blocks.lua -f commonmark -t html sample.md | \
+		$(DIFF) expected.html -
+
+.PHONY: all
+all: test
+
--- a/commonmark-code-blocks/README.md
+++ b/commonmark-code-blocks/README.md
@@ -1,0 +1,23 @@
+# commonmark-code-blocks 
+
+This filter produces code blocks with language CSS class format
+suggested by the CommonMark specification ("language-$lang").
+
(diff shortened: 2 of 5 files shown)


dmbaturin:
Tests do pass in my environment (Fedora 35), so I'll need to find why the diff behavior is different in the CI container. I'll look into it today or tomorrow.

dmbaturin:
Ok, found the issue. The cause was my use of `DIFF ?=` in the Makefile (which I copied from another Makefile in the same repo, which made me think it's a safe thing to do). In reality it wasn't because `DIFF` is set to `diff -u` globally in the container environment, so my options to ignore whitespace changes and trailing newlines were ignored.

I verified that the test passes in the docker test image.

tarleb:
Thank you for this useful filter. We are currently not merging any new filters into this repo, please see #207 for details.

dmbaturin:
@tarleb Maybe it's a good idea to add that information to the README and to CONTRIBUTING.md so that no one tries to open new pull requests.

There's no decision regarding the future structure yet?

Issue #144, opened 2020-11-21
Inkscape deprecated options


twsh:
I am using Inkscape 1.0.1 on Ubuntu 20.10. When I use the (fantastic!) diagram-generator.lua filter to target a format which wants SVG I get the following warnings:

```
"inkscape" --without-gui --file="/tmp/tikz2image-dd825d430cc7d3ac/tikz-image.pdf" --export-plain-svg="/tmp/tikz2image-dd825d430cc7d3ac/tikz-image.svg"
Warning: Option --without-gui= is deprecated
Warning: Option --file= is deprecated
Warning: Option --export-plain-svg= is deprecated

** (org.inkscape.Inkscape:5016): WARNING **: 13:36:25.959: Fonts dir '/usr/share/inkscape/fonts' does not exist and will be ignored.
```
I assume that I can safely ignore these.

Looking at https://inkscape.org/doc/inkscape-man.html I think that the code can be changed to avoid these. I tried this:

```
local function convert_with_inkscape(filetype)
  -- Build the basic Inkscape command for the conversion
  local inkscape_output_args
  if filetype == 'png' then
    inkscape_output_args = '--export-type=png --export-dpi=300'
  elseif filetype == 'svg' then
    inkscape_output_args = '--export-type=svg'
  else
    return nil
  end
  return function (pdf_file, outfile)
    local inkscape_command = string.format(
      '"%s" "%s" ' .. inkscape_output_args .. ' --export-filename="%s"',
      inkscape_path,
      pdf_file,
      outfile
    )
    io.stderr:write(inkscape_command .. '\n')
    local command_output = io.popen(inkscape_command)
    -- TODO: print output when debugging.
    command_output:close()
  end
end
```

PeterSommerlad:
came up independently with a slightly different solution that switches the arguments to format to have the input pdf file last.

```
local function convert_with_inkscape(filetype)
  -- Build the basic Inkscape command for the conversion
  local inkscape_output_args
  if filetype == 'png' then
    inkscape_output_args = '--export-png="%s" --export-dpi=300'
  elseif filetype == 'svg' then
    inkscape_output_args = '--export-plain-svg --export-filename="%s"'
  else
    return nil
  end
  return function (pdf_file, outfile)
    local inkscape_command = string.format(
      '"%s" '  .. inkscape_output_args .. ' "%s" ',
      inkscape_path,
      outfile,
      pdf_file
    )
    io.stderr:write(inkscape_command .. '\n')
[...]

PR #227, opened 2022-05-09
Avoid Inkscape 1.0 deprecation warnings

postylem:
Update the arguments passed to Inkscape, in order to be in line with current Inkscape argument formats (introduced in Inkscape 1.0, released May 2020).

The current arguments passed to Inkscape result in warnings being printed, if the Inkscape version is 1.0 or later. 

----

Updating these arguments avoids the following deprecation warnings (which are safe to ignore, but annoying):
```
Warning: Option --without-gui= is deprecated
Warning: Option --file= is deprecated
```
and
```
Warning: Option --export-plain-svg= is deprecated
```
if the format is svg, or
```
Warning: Option --export-png= is deprecated
```
if the format is png.

### DIFF
--- a/diagram-generator/diagram-generator.lua
+++ b/diagram-generator/diagram-generator.lua
@@ -154,48 +154,30 @@ local function convert_with_inkscape(filetype)
   -- Check inksape version.
   -- TODO: this can be removed if supporting older Inkscape is not important.
   local inkscape_v_string = io.popen(inkscape_path .. " --version"):read()
-  local inkscape_v_major = inkscape_v_string:match(" ([0-9%.]*) "):gmatch("([0-9]*)%.")()
-  if tonumber(inkscape_v_major) < 1 then
-    -- If Inkscape is pre-1.0 version then use old format of arguments
-    if filetype == 'png' then
-      inkscape_output_args = '--export-png="%s" --export-dpi=300'
-    elseif filetype == 'svg' then
-      inkscape_output_args = '--export-plain-svg="%s"'
-    else
-      return nil
-    end
-    return function (pdf_file, outfile)
-      local inkscape_command = string.format(
-        '"%s" --without-gui --file="%s" ' .. inkscape_output_args,
-        inkscape_path,
-        pdf_file,
-        outfile
-      )
-      local command_output = io.popen(inkscape_command)
-      -- TODO: print output when debugging.
-      command_output:close()
-    end
+  local inkscape_v_major = inkscape_v_string:gmatch("([0-9]*)%.")()
+  local isv1 = tonumber(inkscape_v_major) >= 1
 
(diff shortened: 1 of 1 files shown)


postylem:
I just noticed that this has been commented on before in #144, but not implemented in a PR?  This is my first PR here, so I hope I'm being helpful.

alerque (approved):

diagram-generator/diagram-generator.lua
```diff
@@ -150,24 +150,52 @@ local tikz_template = [[
...
   local inkscape_output_args
-  if filetype == 'png' then
-    inkscape_output_args = '--export-type=png --export-dpi=300'
-  elseif filetype == 'svg' then
-    inkscape_output_args = '--export-type=svg --export-plain-svg'
+
+  -- Check inksape version.
+  -- TODO: this can be removed if supporting older Inkscape is not important.
+  local inkscape_v_string = io.popen(inkscape_path .. " --version"):read()
+  local inkscape_v_major = inkscape_v_string:match(" ([0-9%.]*) "):gmatch("([0-9]*)%.")()
```
postylem, line 157: I added this branch so that now it won't break anything if you are using older inkscape.  This can be removed if that is no longer supported.  But for now it probably should be supported.

postylem:
OK, I saw that the test CI is running with Ubuntu 20.04 on a docker container which has Inkscape v0.92.5 installed. So this doesn't pass the test on that setup since the new argument format is not accepted by the old Inkscape.  

I pushed a fix which checks the inkscape version and uses the right arguments depending.

diagram-generator/diagram-generator.lua
alerque, line 157: Can this be done with one call to `:gmatch()`?

alerque (changes requested):
Good catch, I have v1.0 too but given how slow adoption is making this accessible to v0.9 users makes sense. My suggestion to refactor isn't required if you don't like it, just my suggestion for less code and hence easier to read/understand what is happening.

diagram-generator/diagram-generator.lua
postylem: Ah yes, it could.

diagram-generator/diagram-generator.lua
```diff
@@ -150,24 +150,52 @@ local tikz_template = [[
...
+  if tonumber(inkscape_v_major) < 1 then
```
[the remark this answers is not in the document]
postylem: I was going back and forth about something like that earlier.  I think you're right now that I look at it again. A little confusing, but it's better than all the repeated code.

alerque (approved):
Thanks for the refactor, I think that's quite a bit clearer now.

tarleb:
Thanks!

We could avoid re-checking the version for every graphic, but the performance impact should be small. It's probably not a big issue.

PR #231, opened 2022-05-17
shortcodes filter

jjallaire:
Shortcodes are a feature of Quarto implemented as a Lua filter that @nichtich asked us to factor out and contribute here (https://github.com/quarto-dev/quarto-cli/issues/904) so that they can be used outside of Quarto.

Shortcodes are a form of parameterized macro substitution, with syntax based on that already used by [Hugo shortcodes](https://gohugo.io/content-management/shortcodes/). For example, the following shortcode prints the `title` from document metadata:

``` markdown
{{< meta title >}}
```

Shortcodes can return an inline or block element (or list of either) and can be used in both inline and block contexts (the results are normalized as appropriate).

The `shortcodes.lua` filter supports several shortcodes natively:

| Shortcode                                    | Description                            |
|----------------------------------------------|----------------------------------------|
| `meta`                                       | Print value from document metadata     |
| `env`                                        | Print system environment variable      |
| `pagebreak`                                  | Insert a native page-break             |

In addition, you can create custom shortcodes in Lua (see details in [README.md](https://github.com/jjallaire/lua-filters/tree/master/shortcodes#shortcodes)).

### DIFF
--- a/shortcodes/Makefile
+++ b/shortcodes/Makefile
@@ -1,0 +1,14 @@
+DIFF ?= diff --strip-trailing-cr -u
+PANDOC ?= FOO="bar" pandoc
+
+test:
+	@$(PANDOC) --lua-filter=shortcodes.lua --output=test/output.html test/sample.md
+	@$(DIFF) test/expected.html test/output.html
+	@rm -f test/output.html
+	
+expected: test/sample.md shortcodes.lua
+	$(PANDOC) test/sample.md --lua-filter=shortcodes.lua --output test/expected.html
+
+	
+.PHONY: test
+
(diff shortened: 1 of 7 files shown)


jjallaire:
Discussing this w/ @tarleb he pointed out that there is some debate about how these filters should be aggregated (repo w/ embedded source code as is now, list of links, etc.): https://github.com/pandoc/lua-filters/issues/207. I'll close this PR for now pending resolution of that discussion (happy to contribute this in whatever fashion makes after there is more clarity)

tarleb:
For reference: The filter has been published at <https://github.com/quarto-dev/quarto-shortcodes>.

PR #240, opened 2022-07-01
Fix input parsing in cito filter

larsgw:
Fix input parsing when the second part includes a colon.

See https://github.com/manubot/rootstock/issues/420#issuecomment-980077853


tarleb:
Thank you!

PR #247, opened 2022-08-24
fix inkscape command for PNG

twsh:
A space is needed between the commands, and was missing in https://github.com/twsh/lua-filters/commit/aa3fc36504c67dfbf64421ce6c02018a3777ce80


tarleb:
Thanks!

PR #248, opened 2022-09-17
support for mermaidjs [enhancement] v2

garfvl:
Mermaid

EDIT: this PR is based on #181 and adds a couple of fixes to be able to use `mermaid-cli` inside a container (and fix the docker tests).

To use Mermaid, you must install `mermaid-cli` itself. See the
[mermaid-cli website](https://github.com/mermaid-js/mermaid-cli) for more details.

This filter assumes that the `mmdc` command is located in the path
and therefore can be used from any location. Alternatively, you can
set the environment variable `MERMAID` or use the pandoc's meta variable
`mermaid_path`.

Additionally, if you need to add additional parameters to the `mermaid-cli` command, you can use the `MERMAID_OPTIONS` environment variable.

Example usage from [the Mermaid page](https://github.com/mermaid-js/mermaid):

```
sequenceDiagram
Alice->>John: Hello John, how are you?
loop Healthcheck
    John->>John: Fight against hypochondria
end
Note right of John: Rational thoughts!
John-->>Alice: Great!
John->>Bob: How about you?
Bob-->>John: Jolly good!
```

### DIFF
--- a/.tools/Dockerfile
+++ b/.tools/Dockerfile
@@ -15,13 +15,49 @@ RUN apt-get -q --no-allow-insecure-repositories update \
        aspell-en \
        aspell-fr \
        asymptote \
+       ca-certificates \
+       chromium-browser \
        default-jre \
        dvisvgm \
+       fonts-liberation \
        graphviz \
        imagemagick \
        inkscape \
+       libappindicator3-1 \
+       libasound2 \
+       libatk-bridge2.0-0 \
+       libatk1.0-0 \
+       libc6 \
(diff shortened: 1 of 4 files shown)


garfvl:
> Just commenting to let you know that I've got this PR on my todo list; I didn't have time for a review yet.

No problem, thanks for the notification.

tarleb:
The diagram generator is now maintained at https://github.com/pandoc-ext/diagram; I've added Mermaid support there.

Thank you for the contribution!

PR #249, opened 2022-09-21
Add examples, clarify usage in lilypond README 

ddougj:
Examples of the proper use of the lilypond filter could be deduced from
the test files and the Makefile, but not all users would know to
look there. The README has been updated to make this explicit. It
has also been revised to note that the test files can be compiled
with make.

Example conversion commands have also been added to the README,
for the benefit of new users.

The command-line examples contain a note about musical symbols in
running text with PDF output: one’s TeX engine of choice must be
configured to use a typeface that contains such glyphs as Unicode
characters.

See: https://github.com/pandoc/lua-filters/issues/167

### DIFF
--- a/lilypond/README.md
+++ b/lilypond/README.md
@@ -3,7 +3,8 @@ This filter renders [LilyPond](http://lilypond.org) inline code and
 code blocks into embedded images of musical notation. It's designed with
 (pandoc-flavored) Markdown input in mind.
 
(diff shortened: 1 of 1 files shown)


ratnanil:
> The `lilypond` executable must be installed to a location on your `PATH`

Oh sorry, completely missed that one :sleepy:

tarleb:
Thanks, and sorry for the wait.

PR #258, opened 2022-12-21
Notify include-files should be placed before xnos

mattorp:


tarleb:
Thank you, more documentation is always welcome. Maybe we could add a sentence on the *why*, describing how xnos can would only see a code block and not the referenced document.

mattorp:
True, it always help when we build our intuition of what is happening. In hindsight it seems obvious, but if you're not focusing on it, it may never occur that the order could be a problem. 

I believe it didn't cause a reference error, the xnos filters simply didn't take effect. I can verify and update the description later today.

mattorp:
As a newcomer to Pandoc, it's incredible to think about how many powerful tools are accessible before even having to think about how they function. But as with many other tools, you can quickly run into trouble if you don't stop to think or check how they work.

Is there an equivalent to ESLint that can check the order of the filters? Otherwise, I'd like to help look into a solution for this.

tarleb:
That's a good improvement! 

One question: @ilonasilverwood kindly offered help with documentation issues, and she might be able to help us improve this further. Just let us know if you're interested to receive tips on how to improve this further, or if you'd rather see this PR get merged quickly, in which case Ilona and I would pick it up. Either is fine.

mattorp:
Thanks! For me either is fine, just wanted to give a quick suggestion for the description instead of opening an issue. But I think a better description may be the way to go.On 22 Dec 2022, at 16.07, Albert Krewinkel ***@***.***> wrote:﻿
That's a good improvement!
One question: @IlonaSilverwood kindly offered help with documentation issues, and she might be able to help us improve this further. Just let us know if you're interested to receive tips on how to improve this further, or if you'd rather see this PR get merged quickly, in which case Ilona and I would pick it up. Either is fine.

tarleb:
OK, thanks. I'll go ahead and merge then.

tarleb:
aaaand I made a typo in the merge message... sorry, my bad

PR #260, opened 2023-01-06
Changed "prerequisites" to "note" and reworded section content.

IlonaSilverwood:


tarleb:
Great, thanks! It seems that all blocks have been re-wrapped to a new line length, which makes it a bit harder to see the actual changes. Not a big problem in this context, thus merging.

PR #264, opened 2023-03-09
Fixing issue related to pandoc-crossref compatibility

awa5114:
`minted.lua` unconditionally strips the code block identifier when processing with `pandoc-crossref`. This makes the current implementation of `minted.lua`  incompatible with `pandoc-crossref`. Please see [this thread ](https://github.com/lierdakil/pandoc-crossref/issues/379) for more details.

This change allows documents filtered by `pandoc-crossref` and *then* by `minted.lua` to preserve references. This combines the best of both worlds.

I was not able to run the tests to check if this is a breaking change, because of [this issue](https://askubuntu.com/questions/1458474/unable-to-use-apt-in-wsl-ubuntu-due-to-security-issues) in my WSL install. If someone can show me how to run the tests for `minted.lua` on a Windows machine and there are breaking changes I'd be glad to work through them or add new tests if necessary.

@tarleb

### DIFF
--- a/minted/minted.lua
+++ b/minted/minted.lua
@@ -404,12 +404,21 @@ function CodeBlock(block)
   if FORMAT == "beamer" or FORMAT == "latex" then
     local language   = minted_language(block, MintedBlock)
     local attributes = minted_attributes(block, MintedBlock)
+
+    if (block.identifier == "") then
+        label = ""
+    else
+        label = string.format("\\label{%s}", block.identifier)
+    end
+
     local raw_minted = string.format(
-      "\\begin{minted}[%s]{%s}\n%s\n\\end{minted}",
+      "%s\n\\begin{minted}[%s]{%s}\n%s\n\\end{minted}",
+      label,
       attributes,
       language,
       block.text
     )
+
     -- NOTE: prior to pandoc commit 24a0d61, `beamer` cannot be used as the
     -- RawBlock format.  Using `latex` should not cause any problems.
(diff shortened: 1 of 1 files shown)


tarleb:
I don't know much about the minted filter; maybe @svenevs as the author of that filter can help?

svenevs:
Sure, I can try :upside_down_face: I've been out of touch with this stuff for years, wrote that filter during jury duty.  @amine-aboufirass it would be helpful if you could share with me a quick minimal working example of a document that is using both minted and pandoc-crossref.  I know I was using crossref at the time, vaguely remember filter order being a thing and that crossref always has to be first or something.  But I should be able to run the tests this weekend if I can find a bootable linux system that's not my work computer.

Existing tests and were happy: https://github.com/pandoc/lua-filters/actions/runs/4375232713/jobs/7655760573#step:4:52

but maybe we should be adding to the test something that uses crossref?  not sure if that's going to be easy

svenevs:
oh sorry read too quick, https://github.com/lierdakil/pandoc-crossref/issues/379 looks like it has enough for me to work with

badumont:
For your information, the minted package defines \listoflistings exactly as pandoc-crossref does, so it would be safe to reset it to ***@***.*** after minted has been loaded (it would only prevent users from redefining the “List of listings” string via \listoflistingscaption, but pandoc-crossref provides the lolTitle metadata variable for that).

PR #266, opened 2023-03-11
Added support for mermaid (again)

lsignac:
I wrote this before looking at all these PR in lua-filters repo...
My code is a little bit different from this one :   #181  so I decided to make a PR if it can help.
When generating pdf output, pdf should be stretched to the diagram (option -f) or you will get a whole white page with a small diagram inside.
It seems that something is wrong with svg output https://github.com/mermaid-js/mermaid-cli/issues/112
It didn't work correctly with my 10.0.0 mmdc version. I decided to render to pdf and convert pdf to svg (you must have svg2pdf). I think this is a temporary situation.

I read a few lua filters that make a job similar to diagram-generator. Some of them have a more precise way of choosing the output format based on the type of diagram and the pandoc output format.  You can say things like : if the output format is latex or pdf, and the diagram is mermaid, then preferred output file type is pdf or else svg.

Fill free to delete my PR if it does not help.

### DIFF
--- a/diagram-generator/diagram-generator.lua
+++ b/diagram-generator/diagram-generator.lua
@@ -62,6 +62,11 @@ local pdflatex_path = os.getenv("PDFLATEX") or "pdflatex"
 -- "asymptote_path".
 local asymptote_path = os.getenv ("ASYMPTOTE") or "asy"
 
+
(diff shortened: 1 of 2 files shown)


tarleb:
I've added Mermaid support to the filter linked above.

PR #278, opened 2024-04-24
Fix error in pandoc-quotes.lua README

DominikVogel:
The readme redirects to an outdated version of the filter.


tarleb:
Thanks. The latest version of that filter is here: https://github.com/odkr/pandoc-quotes.lua/blob/main/pandoc-quotes.lua

DominikVogel:
Thx. Updated it.

tarleb:
Thanks

